PR #21, opened 2012-05-18
Fixed exception on relative redirects by adding the scheme://host back into the request

jbroadway:
Relative redirects (e.g., `/redirect/here`) were triggering a Requests_Exception "Only HTTP requests
are handled." because it's verifying http(s) is present. This adds the missing scheme up to the
hostname back into the URL before redirecting.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -417,8 +417,8 @@ class Requests {
 				$options['redirected']++;
 				$location = $return->headers['location'];
 				if (strpos ($location, '/') === 0) {
-					// relative redirect, add the scheme://user:pass@hostname back in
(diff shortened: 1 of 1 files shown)


Ugo-PK:
I had the following exception appearing on a particular request, Requests_Exception: Only HTTP requests are handled
It was not really an issue with the lib itself, but I encountered a site that would redirect with a relative URI not starting with with a "/". (eg: index.php?controller=AdminOrders&token=)
To solve the problem I had to change the test for URL absolutize in Requests .php parse_response functions.

`if (strpos ($location, '/') === 0 ) {`

became

`if (strpos ($location, '/') === 0 || strpos ($location, 'http') === false) {`

It would prolly be possible to just absolutize the url without test or add an option to force it.

PR #75, opened 2013-09-24
Add proxy testing

rmccue:
We can use [mitmproxy](http://mitmproxy.org/) to test our proxy support. It's nice and self contained, plus it's scriptable. Requires Python, but I can deal with that (and Travis has it built-in).

Supersedes the remaining issues from #70 and #72.


rmccue:
Looks like HTTP proxies are failing on HHVM, yay for testing!

rmccue:
I think I've worked it out! We're using port 9000, but that happens to be the port that HHVM uses for FastCGI by default. Switching, hopefully that fixes it.

PR #110, opened 2014-04-02
Use select() syscall to avoid busy waiting

diegode:
Currently it consumes a lot of CPU, it's a no-brainer i think.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -156,6 +156,9 @@ class Requests_Transport_cURL implements Requests_Transport {
 		do {
 			$active = false;
 
+			if ($active) {
+				curl_multi_select($multihandle);
+			}
 			do {
 				$status = curl_multi_exec($multihandle, $active);
 			}


TysonAndre:
I have a branch which independently has the exact same approach (to the exact same problem) in a slightly different location. See https://github.com/rmccue/Requests/pull/284

https://secure.php.net/manual/en/function.curl-multi-select.php#refsect1-function.curl-multi-select-returnvalues

>  On failure, this function will return -1 on a select failure (from the underlying select system call).

We probably want to usleep() in the rare case (OS/PHP/curl setup?) that curl_multi_select returns -1, but this is something I'm in favor of.

schlessera:
This PR was closed inadvertently when we changed the stable branch from `master` to `stable`. GitHub was supposed to smoothly adapt the PRs to point to the new branch, but this seems to have failed for the cases where the author's branch/fork of the PR is not available anymore.

jrfnl:
PR recreated as #493

PR #117, opened 2014-05-11
fix overriding port in case of ssl

meox:
I've found this problem using https connection on port different from 443.
I think that this could fix the issue


dd32:
An equivalent change was merged in #237

jrfnl:
Closing as duplicate of already merged PR #237, which was released in version 1.7.0.

PR #123, opened 2014-07-07
Provide cacert option

nicot:
Gives the user the option to provide a path to a certificate authority file.


nicot:
Great, thanks!

PR #124, opened 2014-08-18
Added example on how to use Cookie

catharsisjelly:
Took me a little while to figure out how to use Cookies in the request so I though I would contribute an example


rmccue:
Thanks! You can also pass in a `Requests_Cookie_Jar` as well, which is the preferred option, and it'll add the required headers for you.

(Happy to accept a follow-up PR to add that into the examples too!)

PR #133, opened 2014-12-12
GET request should respect query data parameter

DrColossos:
GET seems to ignore any data that gets passed. It should be able to process query strings for the URL.


rmccue:
Closing in favour of #143/#188.

Issue #142, opened 2015-03-05
Problem with GET parameters 


rootuser:
I am confused; while the requests lib is working great basically, there seems to be a severe issue with GET requests; or maybe I get the concept wrong, for it is not covered in the documentation.

Basically, I wish to execute a GET request with parameters, so it would look like
http://some.uri.tld/some-service?param1=abcde&param2=12345

The 'real' work of requests is done in the Requests_Transport_cURL class; in the function 'setup_handle' the cURL handle is prepared, and there are a few lines regarding the request data:

if (in_array($options['type'], array(Requests::HEAD, Requests::GET, Requests::DELETE)) & !empty($data)) {
			$url = self::format_get($url, $data);
		}
		elseif (!empty($data) && !is_string($data)) {
			$data = http_build_query($data, null, '&');
		}

Thing is: though there obviously is a formatter function to convert the $data parameter array to the GET URI format, the $data variable will always be empty for GET requests, because the request::get function has no $data parameter and passes NULL for $data to the requests::request function.

Can you confirm this being an issue, or am I geting something utterly wrong?

Thank you, 
regards, Tom

ulukaiulukai:
I also just stumbled upon this. Setting the query string of GET requests via the data parameter seems to be broken.

    $data = ['test' => '1'];

    $url = 'http://httpbin.org/get';
    Requests::get($url, [], $data); // returns {"args": {}}

Workaround:

    $url = 'http://httpbin.org/get?' . http_build_query($data);
    Requests::get($url); // returns {"args": {"test": "1"}}

This isn't exactly what the data parameter is supposed to do (because it should also merge the data with an existing query string), but it's close enough.

I wonder why this isn't covered by the tests?

rmccue:
This is an intentional part of the API design; in a typical use case, you won't necessarily need data sent along with a request. Building the URL for you is just a convenience.

`Requests::get` is a helper function designed to make GET requests lightweight in the code, which is why there's no `$data` parameter there. If you need to send data, use `Requests::request` instead:

```
$headers = [];
$data = ['test' => '1'];
$options = [];
$response = Requests::request( 'http://httpbin.org/get', $headers, $data, Requests::GET, $options );
// GET is the default for type, and $options can be blank, so this can be shortened:
$response = Requests::request( 'http://httpbin.org/get', $headers, $data );
```

@ulukaiulukai Your code there is actually setting the `test` option to `'1'`, which doesn't do anything. :) Use `Requests::request` as above instead.

[...]

Issue #91, opened 2013-12-28
DELETE does not allow POST-style data


fluxsauce:
One of our API calls using method DELETE requires JSON in a POST field. We had been doing it with straight cURL. When transitioning to Requests, I tried the delete method but that didn't allow data to be passed. Then, I attempted to just use the request method.

    $response = Requests::request($url, $headers, $data, Requests::DELETE, $options);

URL:

https://REMOVED/terminus.php?site=UUID&path=code-branch

headers:

````
Array (
    [Content-Type] => application/json
    [Content-Length] => 40
)
````

data:

    {"data":{"refspec":"refs\/heads\/prog"}}

method: DELETE

options:

````
Array (
    [timeout] => 30
    [cookies] => Array (
            [0] => REMOVED
        )
    [verifyname] => FALSE
    [verify] => FALSE
)
````

That caused:

    http_build_query(): Parameter 1 expected to be Array or Object. Incorrect value given cURL.php:327

https://github.com/rmccue/Requests/blob/master/library/Requests/Transport/cURL.php#L327

I believe the problem is at:

https://github.com/rmccue/Requests/blob/022400b23f697792fa340c93d250257293c6ec37/library/Requests/Transport/cURL.php#L224

This assumes that DELETE won't be POSTing. Additionally:

https://github.com/rmccue/Requests/blob/master/library/Requests/Transport/cURL.php#L241

cURL isn't setting options for CURLOPT_POST or CURLOPT_POSTFIELDS.

I'll admit that this is a bit edge-case. However, it doesn't seem to be against RFC. http://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request gets into that in greater detail.

I have no problem setting an additional option or calling the request method in a different way.

Let me know if you have any questions or if you need more context! Thanks in advance.

shrivedog:
Currently encountering this behavior/bug in my API testing.

Any decisions on the implementation strategy, or assignment of the coding work?

Personally I rather enjoy the convenience of automatic parsing of the data into a query string, and would opt for a boolean option to indicate when this should be done. Default for the option is a toss up.

LewawiTeshome:
![Uploading Cover .jpg…]()

kadimi:
[Solution](https://github.com/WordPress/Requests/issues/347#issuecomment-531602778) on #347 

> You need to use Requests::request to send GET, DELETE, or OPTIONS requests with bodies, rather than using the convenience methods.

PR #143, opened 2015-03-07
Fix/Rationalize transports

### LINKED ISSUE #137: Support for OPTIONS http method
Could you please add support for OPTIONS http method. We are using this method for feature detection from one of our APIs.

egeloen:
Hey!

This PR rationalizes/fixes some issues I encounter with the curl/fsockopen transports + add some new features.

### Context

I'm currently integrating this library in an http adapter (See egeloen/ivory-http-adapter#80). This library has a pretty robust test suite which have spotted some issues. This PR tries to fix all of them :)

### Issues

 * The curl transport initializes a curl resource at initialization. Then, if this resource is closed, the transport is not usable anymore. Instead of creating it at initialization, I have moved it in the `setup_handle` method which fixes the issue.
 * The curl transport merges the data in the uri for `DELETE` request but some service like AWS needs some extra info in `DELETE` request which needs to be passed by message body. According to #91, I have added the `data_as_query` option and updated the transport accordingly.
 * The fsockopen transport added arbitrary headers without looking for existence in the passed headers. This result of a duplication of headers.

### New features

 * Added support for the `OPTIONS` and `TRACE` http verbs.
 * Added support for `1.0` and `1.1` protocol versions in request/response.

Btw, this PR fixes #133, #137, #142 and #91. 

Waiting your feedbacks!

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -64,6 +64,28 @@ class Requests_Transport_cURL implements Requests_Transport {
 	public function __construct() {
 		$curl = curl_version();
 		$this->version = $curl['version_number'];
+		$this->fp = curl_init();
+
+		curl_setopt($this->fp, CURLOPT_HEADER, false);
+		curl_setopt($this->fp, CURLOPT_RETURNTRANSFER, 1);
+		if ($this->version >= self::CURL_7_10_5) {
+			curl_setopt($this->fp, CURLOPT_ENCODING, '');
+		}
+		if (defined('CURLOPT_PROTOCOLS')) {
+			curl_setopt($this->fp, CURLOPT_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
+		}
+		if (defined('CURLOPT_REDIR_PROTOCOLS')) {
+			curl_setopt($this->fp, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
+		}
+	}
+
+	/**
+	 * Destructor
+	 */
+	public function __destruct() {
+		if (is_resource($this->fp)) {
+			curl_close($this->fp);
+		}
 	}
 
 	/**
@@ -111,7 +133,7 @@ class Requests_Transport_cURL implements Requests_Transport {
 		}
 
 		$this->process_response($response, $options);
-		curl_close($this->fp);
+
 		return $this->headers;
 	}
 
@@ -212,20 +234,6 @@ class Requests_Transport_cURL implements Requests_Transport {
 	 * @param array $options Request options, see {@see Requests::response()} for documentation
 	 */
 	protected function setup_handle($url, $headers, $data, $options) {
-		$this->fp = curl_init();
-
-		curl_setopt($this->fp, CURLOPT_HEADER, false);
-		curl_setopt($this->fp, CURLOPT_RETURNTRANSFER, 1);
-		if ($this->version >= self::CURL_7_10_5) {
-			curl_setopt($this->fp, CURLOPT_ENCODING, '');
-		}
-		if (defined('CURLOPT_PROTOCOLS')) {
-			curl_setopt($this->fp, CURLOPT_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
-		}
-		if (defined('CURLOPT_REDIR_PROTOCOLS')) {
-			curl_setopt($this->fp, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
-		}
-
 		$options['hooks']->dispatch('curl.before_request', array(&$this->fp));
 
 		$headers = Requests::flatten($headers);
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -231,7 +231,18 @@ class Requests_Transport_cURL implements Requests_Transport {
 		$headers = Requests::flatten($headers);
 
 		if (!empty($data)) {
-			if ($options['data_as_query']) {
+			$data_format = $options['data_format'];
+
+			if ($data_format === 'default') {
+				if (in_array($options['type'], array(Requests::HEAD, Requests::GET, Requests::DELETE))) {
+					$data_format = 'query';
+				}
+				else {
+					$data_format = 'body';
+				}
+			}
+
+			if ($data_format === 'query') {
 				$url = self::format_get($url, $data);
 				$data = '';
 			}
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -64,19 +64,6 @@ class Requests_Transport_cURL implements Requests_Transport {
 	public function __construct() {
 		$curl = curl_version();
 		$this->version = $curl['version_number'];
-		$this->fp = curl_init();
-
-		curl_setopt($this->fp, CURLOPT_HEADER, false);
-		curl_setopt($this->fp, CURLOPT_RETURNTRANSFER, 1);
-		if ($this->version >= self::CURL_7_10_5) {
-			curl_setopt($this->fp, CURLOPT_ENCODING, '');
-		}
-		if (defined('CURLOPT_PROTOCOLS')) {
-			curl_setopt($this->fp, CURLOPT_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
-		}
-		if (defined('CURLOPT_REDIR_PROTOCOLS')) {
-			curl_setopt($this->fp, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
-		}
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


egeloen:
@rmccue ping?

ozh:
I think @rmccue will want unit tests before considering a merge, as always :)

egeloen:
Thank you @ozh :) I will add them in the next days!

egeloen:
I'm currently adding tests but I can't get the tests passing for the master branch on my computer. Here, the PHPunit output:

``` bash
PHPUnit 4.6.4 by Sebastian Bergmann and contributors.

Configuration read from /home/gelo/Workspace/Rmccue/Requests/tests/phpunit.xml.dist

....FF....F....F...F..F........................................  63 / 814 (  7%)
............................................................... 126 / 814 ( 15%)
..................................................F...F....F... 189 / 814 ( 23%)
.F...F..F...................................................... 252 / 814 ( 30%)
............................................................... 315 / 814 ( 38%)
....................................F...F...................... 378 / 814 ( 46%)
............................................................... 441 / 814 ( 54%)
............................................................... 504 / 814 ( 61%)
............................................................... 567 / 814 ( 69%)
............................................................... 630 / 814 ( 77%)
............................................................... 693 / 814 ( 85%)
............................................................... 756 / 814 ( 92%)
..........................................................

Time: 6.9 minutes, Memory: 41.25Mb

There were 14 failures:

1) RequestsTest_Auth_Basic::testPOSTUsingInstantiation with data set #0 ('Requests_Transport_fsockopen')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Auth/Basic.php:77

2) RequestsTest_Auth_Basic::testPOSTUsingInstantiation with data set #1 ('Requests_Transport_cURL')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Auth/Basic.php:77

3) RequestsTest_Transport_cURL::testGETWithNestedData
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://httpbin.org/get?test=true&test2%5Btest3%5D=test&test2%5Btest4%5D=test-too'
+'http://httpbin.org/get?test=true&test2[test3]=test&test2[test4]=test-too'

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:71

4) RequestsTest_Transport_cURL::testRawPOST
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:119

5) RequestsTest_Transport_cURL::testRawPUT
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:164

6) RequestsTest_Transport_cURL::testRawPATCH
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:194

7) RequestsTest_Transport_cURL::testMultipleWithDifferingMethods
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:571

8) RequestsTest_Transport_cURL::testMultipleToFile
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:671

9) RequestsTest_Transport_fsockopen::testGETWithNestedData
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://httpbin.org/get?test=true&test2%5Btest3%5D=test&test2%5Btest4%5D=test-too'
+'http://httpbin.org/get?test=true&test2[test3]=test&test2[test4]=test-too'

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:71

10) RequestsTest_Transport_fsockopen::testRawPOST
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:119

11) RequestsTest_Transport_fsockopen::testRawPUT
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:164

12) RequestsTest_Transport_fsockopen::testRawPATCH
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:194

13) RequestsTest_Transport_fsockopen::testMultipleWithDifferingMethods
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:571

14) RequestsTest_Transport_fsockopen::testMultipleToFile
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test'
+''

/home/gelo/Workspace/Rmccue/Requests/tests/Transport/Base.php:671

FAILURES!
Tests: 814, Assertions: 1219, Failures: 14.
```

Majority of the failling tests are related to the data `test` which is sent as `application/x-www-form-urlencoded` which is then part of the form json node instead oe being part of the data node. Do you have any idea why they are green on travis but not on my computer?

ozh:
Yep, noticed the same on my computer, no idea why and honestly I didn't bother. When I submit a PR with new tests, I just make sure **my new tests** are OK :smile:

egeloen:
Thanks for your feedback, I will then rely on travis to be sure my tests are OK :)

ozh:
Actually, I have started looking at it. The problem is that, IMHO, the tests here are doing it wrong: we're testing *httpbin and the values it returns*, not Requests. I suspect that [this line](https://github.com/rmccue/Requests/blob/master/composer.json#L18) explains the discrepancies

ozh:
Additionally,  I'm not even sure what httpbin is supposed to return for instance with test `testRawPOST()` : I get different results with https://www.hurl.it/ and http://requestmaker.com/

egeloen:
Seems related to Runscope/httpbin#178

egeloen:
Tests added! Undortunatelly, httpbin does not allow me to test as much feature as I want. For example, http spec specifies that an OPTIONS request can have a body but httpbin does not deal with such case. Even worse, it does not deal with TRACE request at all.

Feedbacks welcome!

ozh:
@egeloen FYI, regarding tests:  #149

rmccue:
Thanks for the PR, and apologies for the delay in responding!

> The curl transport initializes a curl resource at initialization. Then, if this resource is closed, the transport is not usable anymore. Instead of creating it at initialization, I have moved it in the setup_handle method which fixes the issue.

I think this may break backwards compatibility with people using hooks that occur between `__construct` and `setup_handle`, but I need to look into that more.

What would cause the handle to be closed? I'm assuming the main issue is reusing the transport?

> The curl transport merges the data in the uri for DELETE request but some service like AWS needs some extra info in DELETE request which needs to be passed by message body. According to #91, I have added the data_as_query option and updated the transport accordingly.

I like the concept, but let's make the option a bit more generic. I'm thinking `data_format` with the options of `default` (query args for GET/HEAD/DELETE/OPTIONS, body for PUT/POST), `query`, and `body`. Later, we may want to add `json`, etc as well.

(The original reason we didn't support this was because RFC2616 didn't specify that DELETE et al could take a request body, and the general consensus was to treat them similar to GET requests. RFC7231 clarifies that it's allowed, so happy to add this.)

> The fsockopen transport added arbitrary headers without looking for existence in the passed headers. This result of a duplication of headers.

Good catch.

Further comments inline as well.

library/Requests/Transport/cURL.php
```diff
@@ -225,14 +212,31 @@ public function &get_subrequest_handle($url, $headers, $data, $options) {
...
-			$url = self::format_get($url, $data);
-		}
-		elseif (!empty($data) && !is_string($data)) {
-			$data = http_build_query($data, null, '&');
+
+		if (!empty($data)) {
+			if ($options['data_as_query']) {
+				$url = self::format_get($url, $data);
+				$data = '';
+			} elseif (!is_string($data)) {
```
rmccue: The `elseif` should be on the next line after the closing brace

library/Requests/Transport/cURL.php
```diff
@@ -267,6 +274,7 @@ protected function setup_handle($url, $headers, $data, $options) {
 		curl_setopt($this->fp, CURLOPT_REFERER, $url);
 		curl_setopt($this->fp, CURLOPT_USERAGENT, $options['useragent']);
 		curl_setopt($this->fp, CURLOPT_HTTPHEADER, $headers);
+		curl_setopt($this->fp, CURLOPT_HTTP_VERSION, $options['protocol_version'] == 1.1 ? CURL_HTTP_VERSION_1_1 : CURL_HTTP_VERSION_1_0);
```
rmccue: Strict comparison here please, and it'd be better split into a proper `if` statement instead of the ternary.

library/Requests/Transport/fsockopen.php
```diff
@@ -48,7 +48,8 @@ class Requests_Transport_fsockopen implements Requests_Transport {
 	 * @param array $options Request options, see {@see Requests::response()} for documentation
 	 * @return string Raw HTTP result
 	 */
-	public function request($url, $headers = array(), $data = array(), $options = array()) {
+	public function request($url, $headers = array(), $data = array(), $options = array())
+	{
```
rmccue: The opening brace should be on the same line here

library/Requests/Transport/fsockopen.php
```diff
@@ -89,13 +90,12 @@ public function request($url, $headers = array(), $data = array(), $options = ar
 			}
 
 			stream_context_set_option($context, array('ssl' => $context_options));
-		}
-		else {
+		} else {
```
rmccue: Original formatting here is the preferred option.

library/Requests/Transport/fsockopen.php
```diff
@@ -174,7 +160,12 @@ public function request($url, $headers = array(), $data = array(), $options = ar
 		}
 		$out .= "\r\n";
 
-		$out .= "User-Agent: {$options['useragent']}\r\n";
+		$lowerHeaders = array_change_key_case($headers);
```
rmccue: `$lower_headers` as snake_case instead please. Also didn't realise this function existed, cool. :)

We'll eventually change this to be a proper [CaseInsensitiveDictionary](https://github.com/rmccue/Requests/blob/master/library/Requests/Utility/CaseInsensitiveDictionary.php), but this works nicely in the meantime.

egeloen:
> I think this may break backwards compatibility with people using hooks that occur between __construct and setup_handle, but I need to look into that more.

> What would cause the handle to be closed? I'm assuming the main issue is reusing the transport?

I have change my approach to make it reusable without BC break :) Basically, the issue what the fp is initialized in the constructor and then closed with the first request. Instead of closing it for each request, I have closing it when the transport is destructed.

> I like the concept, but let's make the option a bit more generic. I'm thinking data_format with the options of default (query args for GET/HEAD/DELETE/OPTIONS, body for PUT/POST), query, and body. Later, we may want to add json, etc as well.

Done!

> We'll eventually change this to be a proper CaseInsensitiveDictionary

Done!

ping @rmccue

egeloen:
If all is fine, I will squash my commits before merging.

library/Requests/Transport/fsockopen.php
```diff
@@ -121,62 +121,68 @@ public function request($url, $headers = array(), $data = array(), $options = ar
...
 
+		if ($data_format === 'query') {
+			$path = self::format_get($url_parts, $data);
+			$data = '';
+		}
+		else {
+			$path = self::format_get($url_parts, array());
+		}
+
+		$options['hooks']->dispatch('fsockopen.remote_host_path', array(&$path, $url));
```
egeloen: Not sure about this one. Should I only dispatch it for the `query` data format in order to make it fully BC?

egeloen:
ping @rmccue

rmccue:
@egeloen Apologies for the delay. If anything, I'd actually prefer the commits split out and in separate pull requests, as it makes it easier to manage and merge parts as they're ready.

I'm going to do a full review of this, but it'll also need updating from the latest master in any case.

rmccue:
Thanks again for this @egeloen :) I've updated this with some other small fixes in #188, so closing this one out.

PR #144, opened 2015-03-23
quick fix CurlFile support

Omashu:
<code>$data = http_build_query($data, null, '&');</code> breaks CurlFile object in the array data


rolies106:
Is there any solutions to this? Because I keep failed to upload file when using `CurlFile` because `$data` variable need to be an array object.

PR #148, opened 2015-04-17
Ignore coverage report files

ozh:
Just a simple fix to get a cleaner `git status` without the gazillion coverage files


rmccue:
:heart_eyes: :palm_tree:

PR #149, opened 2015-04-19
Tests discrepancies and an attempt to fix them

ozh:
Requests has a pretty awesome test suite, but several test are unreliable because they give different results on Travis than when run locally.

## What this PR covers

In this PR, all tests pass when run locally. Travis will probably screw them because of the server shim that's used.
```sh
-/planetozh/Requests/tests unreliable_tests
$ phpunit
PHPUnit 4.6.4 by Sebastian Bergmann and contributors.

Configuration read from D:\home\planetozh\Requests\tests\phpunit.xml.dist

..............................................................  62 / 814 (  7%)
.............................................................. 124 / 814 ( 15%)
.............................................................. 186 / 814 ( 22%)
.............................................................. 248 / 814 ( 30%)
.............................................................. 310 / 814 ( 38%)
.............................................................. 372 / 814 ( 45%)
.............................................................. 434 / 814 ( 53%)
.............................................................. 496 / 814 ( 60%)
.............................................................. 558 / 814 ( 68%)
.............................................................. 620 / 814 ( 76%)
.............................................................. 682 / 814 ( 83%)
.............................................................. 744 / 814 ( 91%)
.............................................................. 806 / 814 ( 99%)
........

Time: 2.93 minutes, Memory: 26.75Mb

OK (814 tests, 1195 assertions)

Generating code coverage report in HTML format ... done
```

I think that directives and scripts from `.travis.yml` and the composer file **should not** give different results when running these tests.

## What this PR does not cover

~~There are more [than 260 queries](https://github.com/rmccue/Requests/blob/master/tests/Transport/Base.php#L265-L377) sent to httpbin.org that are IMHO unnecessary: they test more httpbin return values than the behavior of Requests itself. These queries should be rewritten to use a mock just to test return values without issuing an actual external request~~ Added in commit 5, this obviously dramatically improve test suite runtime while preserving the functionality of tests.

Also, `RequestsTest_Transport_Base::setUp()` is run 46*2=92 times, it should be `setUpBeforeClass()` instead, to be run once before all tests run (`setUp()` runs before each and every test of the class)

So....

Thoughts?

### DIFF
--- a/tests/Auth/Basic.php
+++ b/tests/Auth/Basic.php
@@ -74,7 +74,7 @@ class RequestsTest_Auth_Basic extends PHPUnit_Framework_TestCase {
 		$auth = explode(' ', $auth);
 
 		$this->assertEquals(base64_encode('user:passwd'), $auth[1]);
-        $this->assertEquals('test', $result->data);
+		$this->assertEquals('', $result->form->test);
 	}
 
 	/**
--- a/tests/Transport/Base.php
+++ b/tests/Transport/Base.php
@@ -110,7 +110,7 @@ abstract class RequestsTest_Transport_Base extends PHPUnit_Framework_TestCase {
 		$this->assertEquals(200, $request->status_code);
 
 		$result = json_decode($request->body, true);
-		$this->assertEquals('test', $result['data']);
+		$this->assertEquals('', $result['form']['test']);
 	}
 
 	public function testFormPost() {
@@ -155,7 +155,7 @@ abstract class RequestsTest_Transport_Base extends PHPUnit_Framework_TestCase {
 		$this->assertEquals(200, $request->status_code);
 
 		$result = json_decode($request->body, true);
-		$this->assertEquals('test', $result['data']);
+		$this->assertEquals('', $result['form']['test']);
 	}
 
 	public function testFormPUT() {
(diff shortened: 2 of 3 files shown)


rmccue:
There are essentially two problems here:
1. httpbin is a remote request, so it can fall victim to network issues. You have to be online to run these tests as well, which is annoying when developing on a plane (or anywhere without an internet connection).
2. httpbin is, sadly, unreliable. They break existing functionality all the time (Runscope/httpbin#178), which is super annoying.

https://github.com/RequestsPHP/test-server is an attempt to work around this by having a locally-executed server that looks like httpbin. The problem with this is no. 2 above. What we need to do is set up a `tests.requests.ryanmccue.info` that runs that server, while still allow running it locally.

We could also change some of the tests to use mock transports, but the end-to-end testing is pretty valuable, IMO.

ozh:
We're definitely on the same line here: using mocks when possible to avoid external requests, and still testing several stuff end to end because this is what real world looks like.

At least, you should cherry pick these:
* a64942c249b7ff413f481ac4308ea542188a8a5a: rename test & update external URL, hopefully this will help avoiding a couple timeouts
* 75bcdb9cd841563fb56ca9ce719af1b76508a9b1 to mock httpbin requests when it's just about checking a success/failure code (I mean, that's **260** queries saved!)

([in case you forgot how to cherry pick from a PR](https://gist.github.com/ozh/cbce675ba35e0d336cff))

Regarding RequestsPHP/test-server I think it's a veeeery good idea, but I also think it should not introduce differences between using it and using httpbin.org (or a tests.requests.ryan.com for that matter)

rmccue:
Handled in #170, and I'm going to set up a permanent, publicly-available copy of the test server too. Thanks for your work here, and apologies for not getting to it earlier!

PR #154, opened 2015-04-30
Test with PHP 5.6 and 7.0 nightlies

ozh:
Curious about the output.
5.6 is available, 7.0 is an alias for the nightly build until it's released ([doc](http://docs.travis-ci.com/user/languages/php/#PHP-nightly-builds))


ozh:
Not sure why the build on 7 failed but you could add 5.6 to the matrix anyway

rmccue:
Also handled in #170, see also #171 for PHP 7. Thanks again :)

PR #158, opened 2015-05-09
Sets `stream_headers` method to public to allow calling it from other places.

### LINKED ISSUE #157: Make Requests_Transport_cURL::stream_headers accessible
While fixing [this issue in php-vcr](https://github.com/php-vcr/php-vcr/issues/115) I noticed that a [curl header function](https://github.com/rmccue/Requests/blob/master/library/Requests/Transport/cURL.php#L272) references the protected function [Requests_Transport_cURL::stream_headers](https://github.com/rmccue/Requests/blob/master/library/Requests/Transport/cURL.php#L306). 

This works as long as the caller of the callback is in the same "class scope" but not when calling it from somewhere else like php-vcr does. When running the [integration tests](https://travis-ci.org/php-vcr/php-vcr/jobs/61653312) I get the following error:

```
Invalid callback Requests_Transport_cURL::stream_headers, cannot access protected method Requests_Transport_cURL::stream_headers()
```

When I make the method public, it works. Would you like to do this or would you prefer another way to allow for calling this method?

adri:
Fixes #157.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -303,7 +303,7 @@ class Requests_Transport_cURL implements Requests_Transport {
 	 * @param string $headers Header string
 	 * @return integer Length of provided header
 	 */
-	protected function stream_headers($handle, $headers) {
+	public function stream_headers($handle, $headers) {
 		// Why do we do this? cURL will send both the final response and any
 		// interim responses, such as a 100 Continue. We don't need that.
 		// (We may want to keep this somewhere just in case)


adri:
Decreased coverage seems to be related to skipped tests https://travis-ci.org/rmccue/Requests/jobs/61866354. I'm not sure why they are skipped but it shouldn't be related to the changes in this pull request.

adri:
Can I help to have this merged or otherwise solved?

rmccue:
Hmm, is there a specific use case you need this for? It's intended to be used internally only.

adri:
Yeah the use case is described in this [issue on php-vcr](https://github.com/php-vcr/php-vcr/issues/115).

It would be cool to be able to record and play back requests in functional tests using the Requests library.

rmccue:
Aha, gotcha, so you need to be able to call it from php-vcr directly. I don't particularly want to expose it publicly, but I'm guessing there's no real other way of doing this...

adri:
Well there are ways without using `CURLOPT_HEADERFUNCTION` but I don't think it's worth the effort.

rmccue:
¯\_(ツ)_/¯

I don't have a real problem with making it public, so let's merge it :)

adri:
Thanks :-)

Issue #161, opened 2015-06-18
Has HTTPS support been disabled?


beutnagel:
@rmccue, it throws a Request_Exception when using HTTPS urls. Has HTTPS support been disabled?
![image](https://cloud.githubusercontent.com/assets/1915772/8224502/6ae88cdc-1588-11e5-9f55-8dff98c885c1.png)

rmccue:
That regular expression matches `http://` or `https://`. Check the `$url` value that you're passing in, as it seems like it doesn't match either of those.

beutnagel:
Alright, thank you! Got it. 

I just misunderstood the error message, as it says only HTTP requests and figured that https was then not supported.

rmccue:
Yeah, it's because HTTPS is really just `TLS( HTTP )` :)

beutnagel:
Make sense (now). Once again, thank you very much!

ozh:
This said, might make sense to update the error message to "Only HTTP(S) requests are handled"

rmccue:
Indeed, that'd help to clarify it a bit more. :)

ozh:
@beutnagel this means "please make a PR with an updated message" ;)

PR #162, opened 2015-06-19
Updated exception message to specify both http + https

beutnagel:
Updated exception message for when url does not conform to HTTP or
HTTPS to clarify that both HTTP and HTTPS requests are handled.
#161


rmccue:
Thanks! :sparkles: :cake: :palm_tree:

PR #163, opened 2015-06-19
Modify the error $array flatten method types in some circumstances suggest, Proxy will be lost in the request in question, lost 302 and 301 jump in.

shaoqi:
Modify the error $array flatten method types in some circumstances suggest, Proxy will be lost in the request in question, lost 302 and 301 jump in.


rmccue:
I'm not sure what this PR is for, sorry. Closing since it's also out-of-date; if you'd like this merged, please reopen and update the branch.

PR #164, opened 2015-07-10
addex how to install with composer

royopa:


rmccue:
Thanks for this, and apologies for the wait in getting this merged!

Issue #34, opened 2012-11-30
Add ability to pass in a file pointer


sidchilling:
How to post file data / upload file using requests? I can;t find any documentation on this.

Like in `HttpRequest`, we have a method called `addPostFile()`

masiorama:
Not yet available in the last build. I'm interested in it too.

captn3m0:
PHP 5.5 now has CurlFile support: https://secure.php.net/manual/en/class.curlfile.php

fbm-static:
Anyone have a fork making this work?

captn3m0:
I just shifted to guzzle, which is doing really well in terms of development.

fbm-static:
Yeah, but the main selling point of this library was 5.2 compatibility (not my decision!) which Guzzle doesn't offer.

soulseekah:
Related #289

PR #169, opened 2015-09-24
Update cURL.php

andreipetcu:
Allow multipart/form-data on POST

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -232,7 +232,7 @@ class Requests_Transport_cURL implements Requests_Transport {
 			$url = self::format_get($url, $data);
 		}
 		elseif (!empty($data) && !is_string($data)) {
-			$data = http_build_query($data, null, '&');
+			// $data = http_build_query($data, null, '&');
 		}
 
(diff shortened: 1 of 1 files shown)


rmccue:
I'm not sure what your intention is here, but this will break existing functionality.

andreipetcu:
It will allow the use of CurlFile library or to send files via $data['file'] = '@/path/to/file' with multipart/form-data header instead of application/x-www-form-urlencoded.
Basically the library won't permit file upload because of that code line:
$data = http_build_query($data, null, '&');

It should let cURL handle how to send the data instead of forcing it to send with application/x-www-form-urlencoded header.

I've used the library with that line commented out and works. Why do you say it will break existing functionality?

rmccue:
CurlFile or `@/file` support needs to be supported across both cURL and sockets consistently. Changing this will _on the surface_ allow you to send files, but won't work for sockets, so this will break on any servers without cURL.

In addition, passing an array or object directly to cURL rather than handling it ourselves doesn't have the same behaviour; IIRC, there are subtle problems with different versions of cURL here.

See #34.

PR #176, opened 2015-10-06
Fix PHP 7 compatibility

### LINKED ISSUE #171: PHP 7 compatibility
In #170, we're adding PHP 7 to our build matrix, but it's allowed to fail right now, and it does. Let's fix this.

rmccue:
Looks like we're accidentally using `CURLOPT_FILE` still, but it was obsoleted by #147/#172.

There might be further problems too.

Fixes #171.


rmccue:
Huzzah, we have PHP 7 compatibility! :D

PR #185, opened 2015-10-14
Use wrong.host.badssl.com for SSL domain check test

rmccue:
This replaces tv.eurosport.com, which I've never liked, although it is used in other test suites. badssl.com is specifically for SSL testing, and is used in Chrome to test SSL.


ozh:
Neat host, didnt know about it, thanks :)

PR #188, opened 2015-10-26
Fix/Rationalize transports (v2)

### LINKED ISSUE #91: DELETE does not allow POST-style data

### LINKED ISSUE #137: Support for OPTIONS http method

rmccue:
Following on from @egeloen's fantastic work in #143, this merges master in and corrects some small issues.

Fixes #137, fixes #91.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -64,6 +64,28 @@ class Requests_Transport_cURL implements Requests_Transport {
 	public function __construct() {
 		$curl = curl_version();
 		$this->version = $curl['version_number'];
+		$this->fp = curl_init();
+
+		curl_setopt($this->fp, CURLOPT_HEADER, false);
+		curl_setopt($this->fp, CURLOPT_RETURNTRANSFER, 1);
+		if ($this->version >= self::CURL_7_10_5) {
+			curl_setopt($this->fp, CURLOPT_ENCODING, '');
+		}
+		if (defined('CURLOPT_PROTOCOLS')) {
+			curl_setopt($this->fp, CURLOPT_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
+		}
+		if (defined('CURLOPT_REDIR_PROTOCOLS')) {
+			curl_setopt($this->fp, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
+		}
+	}
+
+	/**
+	 * Destructor
+	 */
+	public function __destruct() {
+		if (is_resource($this->fp)) {
+			curl_close($this->fp);
+		}
 	}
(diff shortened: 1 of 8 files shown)


library/Requests/Transport/cURL.php
```diff
@@ -296,15 +311,18 @@ protected function setup_handle($url, $headers, $data, $options) {
...
+			case Requests::OPTIONS:
 				curl_setopt($this->handle, CURLOPT_CUSTOMREQUEST, $options['type']);
 				curl_setopt($this->handle, CURLOPT_POSTFIELDS, $data);
 				break;
-			case Requests::DELETE:
-				curl_setopt($this->handle, CURLOPT_CUSTOMREQUEST, 'DELETE');
-				break;
 			case Requests::HEAD:
+				curl_setopt($this->handle, CURLOPT_CUSTOMREQUEST, $options['type']);
 				curl_setopt($this->handle, CURLOPT_NOBODY, true);
```
rmccue: @egeloen Looks like `CURLOPT_POSTFIELDS` doesn't get set for HEAD requests; was this intentional?

PR #190, opened 2015-10-26
Allow passing a stream as data input

### LINKED ISSUE #34: Add ability to pass in a file pointer

rmccue:
Fixes #34. Still in progress.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -572,6 +572,16 @@ class Requests {
 				$options['data_format'] = 'body';
 			}
 		}
+
+		// Check the data stream is valid
+		if (is_resource($data)) {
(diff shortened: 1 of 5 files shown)


rmccue:
The error here on HHVM is facebook/hhvm#4036; need to think of a creative workaround.

laoheimao:
Can I use this  library in YII2 framework?

mediaformat:
Was HHVM the only blocker here?

Trying to get some movement for [PSR-7 support](https://github.com/WordPress/Requests/issues/320)

Issue #191, opened 2015-10-28
Adding a "cURL Info" hook?


brandonhesse:
I'm using this library for a project where I would like to at the time cURL is parsed for it's info via `curl_getinfo` that we have a hook to retrieve that data and store it more locally. Such info like the `size_upload` and `size_download` are only available here, and parsing the body is not valid for gzipped returns which are parsed before being returned with the `Request_Response`.

At the current time in order to parse and retrieve this data from a consumer's end we would have to access it by capturing the curl resource pointer from the `curl.before_send`, and then consume it in a similar manner sometime after the `curl.after_request` event.

My current method:
```php
<?php
// Assign Hooks inside a class
$this->hooks->register('curl.before_send', function($fp){
    $this->curlFp = &$fp;
});
$this->hooks->register('curl.after_request', function($resource) {
    $this->curlInfo = curl_getinfo($this->curlFp);
});
```

Around [line 390](https://github.com/rmccue/Requests/blob/master/library/Requests/Transport/cURL.php#L390) would it be possible to add a hook to send this info on the dispatch?

```php
<?php
//...
if (curl_errno($this->fp)) {
    throw new Requests_Exception('cURL error ' . curl_errno($this->fp) . ': ' . curl_error($this->fp), 'curlerror', $this->fp);
    return;
}
$this->info = curl_getinfo($this->fp);

$options['hooks']->dispatch('curl.curl_info', array($this->info)); //new
$options['hooks']->dispatch('curl.after_request', array(&$this->headers));
return $this->headers;
```

rmccue:
I think we can probably handle this by passing `$this->info` into `$this->headers`

brandonhesse:
That would also work great. It's purely because I wasn't sure how to add it without modifying the current api. Passing it to after_request would be the ideal place to grab it

PR #192, opened 2015-10-29
Return the $info from the transport in the <transport>.after_request event

brandonhesse:
Handles issue #191 
- Returns $this->info from the current $transport on blocking transports only
- Updated documentation to show the new API to the hooks
- Added a test case to validate that it is used at least once and that it's a non-empty array.

I couldn't get phpunit to run on the test, it threw errors about autoloading the base test files. I did download and include the phpunit.phar, but I utilized most of the test from the progress test a few lines above, so it should not have any issues.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -45,7 +45,9 @@ Available Hooks
 
     Alter the raw HTTP response before returning for parsing
 
-    Parameters: `string &$response`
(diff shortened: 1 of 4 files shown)


rmccue:
Thanks for the PR, and sorry for not getting to this sooner. I'm going to update this to fix the tests in a separate PR :)

PR #193, opened 2015-11-07
Create form-auth.php

anarkopsykotik:
I think an example of form authentication; using different objects and explaining why and how to use them would help clarify the documentation.


rmccue:
This is handy if you need to do it, but is usually not the best way to handle authentication. You're better off using Basic auth or OAuth if available on the site than form authentication.

Going to close the PR, but the code is always here if someone comes searching for it :)

PR #194, opened 2015-11-10
RFC / WIP (do not merge): surface cURL errors

Stelminator:
This is a work in progress, but I wanted to get comments sooner rather than later, since this is mostly complete.

I have a couple use cases involved here:

1. I have a need to detect different kinds of cURL errors from multi-requests, specifically CURLE_OPERATION_TIMEDOUT (28), since I have subrequests using timeouts.  These currently manifest as 'requests.no_crlf_separator' due to nothing being returned from curl_multi_getcontent in request_multiple resulting in a parse error.  These are detectable from the code in $done['code'], but none of this is exposed anywhere I can get to.
1. I want to be able to capture the time when an error has finished, and have the code available there.  I've added this as a separate hook to not break existing parse_response implementations, but I'm open to suggestions here.

Please take a look and tell me what you think.

### DIFF
--- a/library/Requests/Exception/Transport/cURL.php
+++ b/library/Requests/Exception/Transport/cURL.php
@@ -1,0 +1,56 @@
+<?php
+
+class Requests_Exception_Transport_cURL extends Requests_Exception_Transport {
+
+	const EASY = 'cURLEasy';
+	const MULTI = 'cURLMulti';
+	const SHARE = 'cURLShare';
+
+	/**
+	 * cURL error code
+	 *
+	 * @var integer
+	 */
+	protected $code = -1;
+
+	/**
+	 * Which type of cURL error
+	 *
+	 * EASY|MULTI|SHARE
+	 *
+	 * @var string
+	 */
+	protected $type = 'Unknown';
+
+	/**
+	 * Clear text error message
+	 *
+	 * @var string
+	 */
+	protected $reason = 'Unknown';
+
+	public function __construct($message, $type, $data = null, $code = 0) {
+		if ($type !== null) {
+			$this->type = $type;
+		}
+
+		if ($code !== null) {
+			$this->code = $code;
+		}
+
+		if ($message !== null) {
+			$this->reason = $message;
+		}
+
+		$message = sprintf('%d %s', $this->code, $this->reason);
+		parent::__construct($message, $this->type, $data, $this->code);
+	}
+
+	/**
+	 * Get the error message
+	 */
+	public function getReason() {
+		return $this->reason;
+	}
+
+}
--- a/library/Requests/Exception/Transport.php
+++ b/library/Requests/Exception/Transport.php
@@ -1,0 +1,5 @@
+<?php
+
+class Requests_Exception_Transport extends Requests_Exception {
+
+}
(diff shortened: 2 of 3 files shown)


library/Requests/Exception/Transport/cURL/Share/5.php
```diff
@@ -0,0 +1,27 @@
...
+     * @var integer
+     */
+    protected $code = 5;
+
+    /**
+     * Reason phrase
+     *
+     * @var string
+     */
+    protected $reason = 'The requested sharing could not be done because the library you use don't have that particular feature enabled. (Added in 7.23.0)';
```
mishan: need to handle apostrophes

mishan:
Pretty cool otherwise :+1:

mishan:
Class 'Requests_Exception_Transport_cURL_Easy' not found in /home/travis/build/rmccue/Requests/library/Requests/Transport/cURL.php on line 234

Stelminator:
yup, only "mostly" complete.  Still looking at what I want that inheritance chain to look like.

rmccue:
I'd think these should all be in a single exception class, `Requests_Exception_Transport_cURL`. Unlike HTTP errors, these are mostly an internal thing. If we can even get the same behaviour across both cURL and sockets, I'd be in favour of a single class there. (We can pass the raw error data including the cURL code itself through in the exception data.)

Generally though, I think the approach of using a separate hook is good here, and we should definitely be detecting the errors better than we do currently. Thanks for your work on the PR so far :palm_tree:

rmccue:
> If we can even get the same behaviour across both cURL and sockets, I'd be in favour of a single class there.

Forgot to mention also: if we can get these consistent across both, I wouldn't be against generic `Requests_Exception_Timeout`/etc exceptions. It's just the internal cURL behaviour I don't want to expose.

Stelminator:
Ah, ok, yeah, as long as I can make a decision on how to handle the error code, I'd be in favor of fewer classes.

Stelminator:
I know this needs some whitespace cleanup, but I can't seem to get the tests to pass locally on master, so I'm not what is and what is not a legitimate failure.

Stelminator:
a bit cleaner now.

Stelminator:
working on fixing up the tests.

Stelminator:
ok, not sure what's up with those last two errors.

Stelminator:
ok, please take a look at this version.  I'm not sure why the tests are failing, but I don't see how the ones remaining could be due to my change.  Earlier, I had messed up a merge/rebase, and so there were legitimate errors.  seems like there's a network issue with the tests.

rmccue:
Sorry for the delay here, had to work out why the tests were failing! Can you merge master into here please? :)

library/Requests/Exception/Transport/cURL.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+class Requests_Exception_Transport_cURL extends Requests_Exception_Transport {
+
+    const EASY = 'cURLEasy';
```
rmccue: Indentation needs to be tabs, not spaces.

library/Requests/Exception/Transport/cURL.php
```diff
@@ -0,0 +1,57 @@
...
+    protected $code = -1;
+
+    /**
+     * Which type of cURL error
+     *
+     * EASY|MULTI|SHARE
+     *
+     * @var string
+     */
+
```
rmccue: Extra newline here needs removing.

Stelminator:
Should be a bit cleaner now.

rmccue:
Fantastic, thanks so much!

PR #196, opened 2015-11-20
Remove sni.velox.ch from the test suite

rmccue:
Oops! :speak_no_evil:


rmccue:
(Only proxy tests failing; will fix those when I get time.)

PR #200, opened 2016-01-13
Fix for memleak and curl_close() never being called

kwuerl:
Curl spammed the file descriptor list with CLOSE_WAIT connections and the Requests_Transport_cURL Instances where never actually destroyed because of a $this reference in the curl handle.
This resulted in a memleak and failing requests after a few 100 being made.
I fixed this by removing the reference from the file handle again after each request.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -176,6 +176,11 @@ class Requests_Transport_cURL implements Requests_Transport {
 
 		$this->process_response($response, $options);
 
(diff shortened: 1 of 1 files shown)


rmccue:
Thanks for the PR! Can you please merge master into this branch? (There was an issue with the tests that I just fixed.)

kwuerl:
Done

rmccue:
Thanks, and apologies for the wait!

PR #207, opened 2016-04-19
Uppercase the method to ensure compatibility

### LINKED ISSUE #204: Small caps http verb don't work as expected
Hello,

I was using request specific function `post()/get()` but then I decided to shift to `request()` but I mistakenly forgot to uppercase the HTTP verbs and type was going as `post/get` which was not recognised by `cURL Transport`.
I think `post` should work same as `POST`, it should be case insensitive.

rmccue:
Easy fix to improve usability. :) Fixes #204.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -564,6 +564,9 @@ class Requests {
 			$url = $iri->uri;
 		}
 
+		// Massage the type to ensure we support it.
+		$type = strtoupper($type);
+
 		if (!isset($options['data_format'])) {
 			if (in_array($type, array(self::HEAD, self::GET, self::DELETE))) {
 				$options['data_format'] = 'query';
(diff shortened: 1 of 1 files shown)


library/Requests.php
```diff
@@ -572,6 +572,9 @@ protected static function set_defaults(&$url, &$headers, &$data, &$type, &$optio
 				$options['data_format'] = 'body';
 			}
 		}
+
```
73sl4: `$type` var should be changed to uppercase at the start of this function since `$type` var is used to define data format.
Let me know if you want me to create PR or update it.

rmccue: Good catch!

73sl4:
@rmccue I think we should also change `$type` in `request_multiple()` since `$request['options']['type']` is set before the call of `set_defaults()`.

rmccue:
@73sl4 It operates by reference, so that'll get changed anyway; the logic in `request_multiple()` is just for merging the "global" options into the per-request options.

73sl4:
@rmccue `set_defaults()` is only uppercasing `$type` var.
`request_multiple()` is setting `$request['options']['type']` which is not changed to uppercase.

PR #211, opened 2016-05-13
Force closing keep-alive connections on old cURL

rmccue:
cURL <7.22 seems to hold the request open if the connection is
keep-alive and the buffer hasn't been filled. Changing to single
connections is slightly less efficient (although unnoticable in most
uses), but should fix this issue.

If you need keep-alive, set the Connection header manually.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -309,6 +309,11 @@ class Requests_Transport_cURL implements Requests_Transport {
 	protected function setup_handle($url, $headers, $data, $options) {
 		$options['hooks']->dispatch('curl.before_request', array(&$this->handle));
 
+		// Force closing the connection for old versions of cURL (<7.22).
+		if ( ! isset( $headers['Connection'] ) ) {
+			$headers['Connection'] = 'close';
+		}
+
 		$headers = Requests::flatten($headers);
 
 		if (!empty($data)) {


schlessera:
@rmccue Do you remember any additional context here around the specific version number you've identified?

I'm asking because actively making use of that specific version for conditional code ended up breaking a lot of things. See https://github.com/WordPress/Requests/issues/838 & https://core.trac.wordpress.org/ticket/59842

rmccue:
@schlessera Don't recall the context off the top of my head.

I checked the cURL changelog for 7.22 and not seeing anything super relevant around keepalives, best I can see is:
> when sending a request and an error is received before the (entire) request body is sent, stop sending the request and close the connection after having received the entire response. This is equally true if an Expect: 100-continue header was used.
> fixed several memory leaks in OOM situations 
> bad expire(0) caused multi_socket API to hang 

Looks like 7.21.7 did have this though:
> Force connection close for HTTP 200 OK when time condition matched 

Other issues I can see that could be relevant are https://github.com/jonashaag/bjoern/issues/51 and [this Stack Overflow question](https://stackoverflow.com/questions/24088035/libcurl-7-21-7-long-timeout-on-curlopt-post-after-tcp-connections-closed-by-the) which links https://stackoverflow.com/questions/10285700/curl-error-recv-failure-connection-reset-by-peer-php-curl

Sounds like in any case based on #838, keepalive may be more generally _still_ be broken, either with cURL in PHP or with w.org.

PR #223, opened 2016-07-12
Support a Requests_Proxy_HTTP() instance for the proxy setting.

ocean90:
See https://core.trac.wordpress.org/attachment/ticket/37107/37107.diff and https://github.com/rmccue/Requests/blob/2781cddf860322d5441362faf08ebc533619dbcd/library/Requests.php#L325-L326.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -323,7 +323,7 @@ class Requests {
 	 *    for Basic authentication
 	 *    (Requests_Auth|array|boolean, default: false)
 	 * - `proxy`: Proxy details to use for proxy by-passing and authentication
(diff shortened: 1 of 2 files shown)


rmccue:
@ocean90 Can you fix the unit test for this too please? [`RequestsTest_Proxy_HTTP::testConnectWithInstance`](https://github.com/rmccue/Requests/blob/2781cddf860322d5441362faf08ebc533619dbcd/tests/Proxy/HTTP.php#L75-L90) should have been testing this, but looks like a copypasta issue.

ocean90:
@rmccue On it.

rmccue:
Wonderful, thanks!

PR #224, opened 2016-07-13
Remove the executable bit

ocean90:
There are a few files that have set the executable bit, let's remove it.


rmccue:
#windowslife

PR #227, opened 2016-07-28
cURL: Support custom HTTP methods

ocean90:
See https://core.trac.wordpress.org/ticket/37503 for background.

The test doesn't work (yet) because there is no /purge endpoint.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -333,13 +333,6 @@ class Requests_Transport_cURL implements Requests_Transport {
 				curl_setopt($this->handle, CURLOPT_POST, true);
 				curl_setopt($this->handle, CURLOPT_POSTFIELDS, $data);
 				break;
-			case Requests::PATCH:
-			case Requests::PUT:
-			case Requests::DELETE:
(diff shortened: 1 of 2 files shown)


rmccue:
https://github.com/RequestsPHP/test-server is the source for the test server. :)

ocean90:
@rmccue No idea why it returns a 501 error. :(

ocean90:
Instead of adding a new case we can also combine the default case with PATCH, PUT, DELETE and OPTIONS. See https://core.trac.wordpress.org/attachment/ticket/37503/37503.2.patch.

@rmccue Let me know if this is a better approach, happy to refresh the PR. (It would also fix the coverage thing. :))

rmccue:
Coverage can be pretty safely ignored for <0.1%; I tried this PR locally and it worked, but not on Travis, so need to get to the bottom of that. Haven't had time yet, alas.

I'd prefer the switch be explicit in what it supports, but `case ...: default:` is fine to avoid the duplication and still be explicit.

rmccue:
Tests passing, wunderbar!

PR #230, opened 2016-08-17
Bail from cURL when either `curl_init()` OR `curl_exec()` are unavailable

dd32:
Some hosts disable cURL by adding `disable_functions = curl_exec` to their `php.ini` file without also adding `curl_init` to the same declaration.
Request's logic appears incorrect here, bailing only if BOTH `curl_init()` AND `curl_exec()` are disabled.


rmccue:
Good catch!

PR #236, opened 2016-08-26
Fixes bug in rfc2616 #3.6.1 implementation.

stephenharris:
A chunk size must be followed by either optional chunk extension(s) (which begin with a semicolon) or \r\n. The current implementation just allows anything from the chunk size to the \r\n.

This was causing a bug in my [Events](http://wordpress.org/plugins/event-organiser/) plugin: a Google iCal feed was *sometimes* mistaken for being chunked (Google claims it is in the header: it isn't).

The starting line is 

```
BEGIN:VCALENDAR
```

which is mistakenly interpreted as a chunk header with size `BE`. *If*, it just so happens the rest of the feed is interpreted as chunked you end up with a mutilated iCal feed.

Of course, Google shouldn't claim it is chunked, but the current implementation fails to detect that it is not a valid encoding.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -749,7 +749,7 @@ class Requests {
 	 * @return string Decoded body
 	 */
 	protected static function decode_chunked($data) {
-		if (!preg_match('/^([0-9a-f]+)[^\r\n]*\r\n/i', trim($data))) {
+		if (!preg_match('/^([0-9a-f]+)(?:;[^\r\n]*)*\r\n/i', trim($data))) {
 			return $data;
 		}
 
@@ -757,7 +757,7 @@ class Requests {
 		$encoded = $data;
 
 		while (true) {
-			$is_chunked = (bool) preg_match('/^([0-9a-f]+)[^\r\n]*\r\n/i', $encoded, $matches);
+			$is_chunked = (bool) preg_match('/^([0-9a-f]+)(?:;[^\r\n]*)*\r\n/i', $encoded, $matches);
 			if (!$is_chunked) {
 				// Looks like it's not chunked after all
 				return $data;
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -749,15 +749,17 @@ class Requests {
 	 * @return string Decoded body
 	 */
 	protected static function decode_chunked($data) {
-		if (!preg_match('/^([0-9a-f]+)(?:;[^\r\n]*)*\r\n/i', trim($data))) {
+		if (!preg_match('/^([0-9a-f]+)(?:;(?:[\w-]*)(?:=(?:(?:[\w-]*)*|"(?:[^\r\n])*"))?)*\r\n/i', trim($data))) {
 			return $data;
 		}
+
+
 
 		$decoded = '';
 		$encoded = $data;
 
 		while (true) {
-			$is_chunked = (bool) preg_match('/^([0-9a-f]+)(?:;[^\r\n]*)*\r\n/i', $encoded, $matches);
+			$is_chunked = (bool) preg_match('/^([0-9a-f]+)(?:;(?:[\w-]*)(?:=(?:(?:[\w-]*)*|"(?:[^\r\n])*"))?)*\r\n/i', $encoded, $matches);
 			if (!$is_chunked) {
 				// Looks like it's not chunked after all
 				return $data;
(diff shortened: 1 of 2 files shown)


stephenharris:
I've updated the test to provide an example of a false-positive that were were previously missing. 

I've also added further test (data) for correct decoding of chunks with chunk extensions.

tests/ChunkedEncoding.php
```diff
@@ -39,7 +43,7 @@ public function testChunked($body, $expected){
 	 */
 	public function testNotActuallyChunked() {
 		$transport = new MockTransport();
-		$transport->body = 'Hello! This is a non-chunked response!';
+		$transport->body = "Believe me\r\nthis looks chunked, but it isn't.";
```
rmccue: Can this be split into a separate test please? Ideally, it should look more like a chunked response (i.e. `2Anotchunked\r\n...`) so it's obvious what it's testing.

rmccue:
Great catch :) One small thing to fix in the tests, but otherwise looks good.

tests/ChunkedEncoding.php
stephenharris: Would a data provider be suitable? The test `testNotActuallyChunked()` hasn't really changed, just the example's we're feeding it.

rmccue: Good point, works for me :)

tests/ChunkedEncoding.php
```diff
@@ -15,6 +15,10 @@ public static function chunkedProvider() {
 				"02\r\nab\r\n04\r\nra\nc\r\n06\r\nadabra\r\n0c\r\n\nall we got\n",
 				"abra\ncadabra\nall we got\n"
 			),
+			array(
+				"02;foo=bar;hello=world\r\nab\r\n04;foo=baz\r\nra\nc\r\n06;justfoo\r\nadabra\r\n0c\r\n\nall we got\n",
```
rmccue: Is OWS (optional whitespace) allowed here too? Should add a test for that if so. Ditto if params can take quoted values.

stephenharris: RFC2616 states:

 > chunk-ext-name = token
 > chunk-ext-val  = token | quoted-string

Token cannot contain any white space, or any of the following: `(}|<>@,;:\"/[]?={}`. Quoted string can contain white space. 

What we have at the moment is still a relatively low-bar for what is interpreted of correctly encoded. I'm wary about making it _too_ strict as a misunderstanding here, when it's too strict, is more likely to lead to issues in production.

Happy to add the above restrictions with some additional tests

library/Requests.php
```diff
@@ -749,15 +749,17 @@ public static function parse_multiple(&$response, $request) {
 	 * @return string Decoded body
 	 */
 	protected static function decode_chunked($data) {
-		if (!preg_match('/^([0-9a-f]+)(?:;[^\r\n]*)*\r\n/i', trim($data))) {
+		if (!preg_match('/^([0-9a-f]+)(?:;(?:[\w-]*)(?:=(?:(?:[\w-]*)*|"(?:[^\r\n])*"))?)*\r\n/i', trim($data))) {
```
stephenharris: I'm not convinced that `[\w-]*` isn't too strict for a token. I would be tempted to err on the side of caution and stick with the original patch and have a slightly lower bar for what constitutes a valid encoding.

rmccue:
Turns out the tests failing here were just a transient issue on a HTTPS connection; restarted the failing build and everything passed nicely. Thanks!

westi:
> a Google iCal feed was sometimes mistaken for being chunked (Google claims it is in the header: it isn't).

Google isn't lying and this change doesn't really fix the issue it just fixes a symptom.

Depending on the underlying transport you may get the chunks already decoded which means chunked decoding can't be keyed directly off the header.

By default `curl` decodes chunks internally and so Requests shouldn't be trying to decode chunks again because it will potentially damage the transferred content.

You can see this with the `curl` cli by using the `--raw` option to disable the decoding when looking at the Google iCal feeds.

rmccue:
@westi Yup, I figured this was the case, but the PR fixes it regardless.

westi:
Except... if the data being transferred "looks" like chunked encoded data it will still mangle it... and we _know_ we don't need to decode so running a bunch of Regex is wasteful

rmccue:
I don't disagree, so happy to accept a PR :)

PR #240, opened 2016-09-18
Allow paths starting with a double-slash

### LINKED ISSUE #231: URLs with double slash at the start of the path fail to be requested
`Requests_IRI` currently rejects a url such as `http://example.org//path/` as being valid, but accepts `http://example.org/path//path/`.

This seems to stem from this bit of code: https://github.com/rmccue/Requests/blob/master/library/Requests/IRI.php#L691-L694

I haven't looked at it close enough to determine if it's a logic error in the if statement, or if it's an edgecase it's specifically not attempting to support. At first I thought it was trying to catch the case where the scheme separator was picked up as the path component.

This [was reported](https://wordpress.org/support/topic/wp_remote_get-is-not-working-after-wordpress-update-to-46) as a regression in WordPress 4.6, as we previously handled this scenario before we switched to Requests.

rmccue:
Fixes #231.

### DIFF
--- a/library/Requests/IRI.php
+++ b/library/Requests/IRI.php
@@ -688,10 +688,7 @@ class Requests_IRI {
 		$isauthority = $this->iuserinfo !== null || $this->ihost !== null || $this->port !== null;
 		if ($this->ipath !== '' &&
 			(
-				$isauthority && (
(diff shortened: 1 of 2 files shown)


cedev935:
Hi, I request to commit.

PR #241, opened 2016-09-22
Make Requests_Transport_cURL::stream_body() public

justinstern:
Fixes a "cannot access protected method Requests_Transport_cURL::stream_body()" error when using PHPUnit and PHP-VCR.

Related issues: #157 and php-vcr/php-vcr#109


rmccue:
Works for me!

justinstern:
@rmccue thanks for the speedy service sir!

PR #249, opened 2016-10-21
Add Content-* headers for empty POST requests

### LINKED ISSUE #248: `Content-Length` header not specified for empty POST requests with fsockopen transport
A reported to <a href="https://core.trac.wordpress.org/ticket/38427">WordPress.org Trac in #38427</a> the fsockopen transport fails to send a `Content-Length: 0` header for empty payloads.

This causes some servers to respond with a 411 error.
To quote the above ticket:

> HTTP result code 411 is "Length required". RFC2616 Section 14.13 says Content-Length header "SHOULD" be included. In practice, it is not used for GET nor HEAD requests, but is expected for POST requests.

dd32:
Sends the `Content-Length` and `Content-Type` headers even for empty POST requests.

The test passes regardless, as it seems that the Heroku test app adds the `Content-Length` headers on incoming requests regardless of if it's sent. Testing against `http://httpbin.org/post` shows that the header isn't/is sent though.

Fixes #248

### DIFF
--- a/library/Requests/Transport/fsockopen.php
+++ b/library/Requests/Transport/fsockopen.php
@@ -160,7 +160,7 @@ class Requests_Transport_fsockopen implements Requests_Transport {
 				$request_body = $data;
 			}
 
-			if (!empty($data)) {
+			if (!empty($data) || $options['type'] === Requests::POST) {
 				if (!isset($case_insensitive_headers['Content-Length'])) {
 					$headers['Content-Length'] = strlen($request_body);
 				}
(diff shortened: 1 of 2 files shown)


dd32:
I'm unable to rebase this as I no longer have the original repository it was made in.

It looks like a new PR combining this & #318 would make sense.

jrfnl:
@dd32 Thanks for letting me know. I've created PR #368 now which consolidates both previous PRs. Feedback welcome.

PR #262, opened 2017-02-08
Run tests with php 7.1 as well, php 7.1.0 was released a month ago

TysonAndre:
The files in this repo are syntactically valid in php, and there don't seem to have been breaking changes in curl

https://github.com/php/php-src/blob/f352c50026e7b9e6116def1c201455f376e5c1a9/NEWS#L376

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ matrix:
     - php: 5.6
       env: TEST_COVERAGE=1
     - php: 7.0
+    - php: 7.1
     - php: hhvm
 
 # Use new container infrastructure


ozh:
@rmccue For the record, if you're investigating this problem: https://github.com/YOURLS/YOURLS/pull/2144#issuecomment-294189704

ozh:
@rmccue Also for the record, if you're into this problem, [my thoughts on how to deal with this](https://github.com/YOURLS/YOURLS-unit-tests/commit/6b6d7607fd953da228b07743b2f296c3380bd79a), HTH.

TysonAndre:
Moving this into #279

PR #263, opened 2017-02-08
Nits on phpdoc, use of internal methods with unexpected type

TysonAndre:
detected by phan

### DIFF
--- a/library/Requests/Cookie/Jar.php
+++ b/library/Requests/Cookie/Jar.php
@@ -68,7 +68,7 @@ class Requests_Cookie_Jar implements ArrayAccess, IteratorAggregate {
 	 * Get the value for the item
 	 *
 	 * @param string $key Item key
-	 * @return string Item value
+	 * @return string|null Item value (null if offsetExists is false)
 	 */
 	public function offsetGet($key) {
 		if (!isset($this->cookies[$key])) {
--- a/library/Requests/IRI.php
+++ b/library/Requests/IRI.php
@@ -67,28 +67,28 @@ class Requests_IRI {
 	/**
 	 * Scheme
 	 *
-	 * @var string
+	 * @var string|null
 	 */
 	protected $scheme = null;
 
 	/**
 	 * User Information
 	 *
-	 * @var string
+	 * @var string|null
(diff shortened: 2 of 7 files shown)


library/Requests/Cookie/Jar.php
```diff
@@ -68,7 +68,7 @@ public function offsetExists($key) {
 	 * Get the value for the item
 	 *
 	 * @param string $key Item key
-	 * @return string Item value
+	 * @return string|null Item value (Not null when offsetExists is true)
```
rmccue: This reads backwards to me; should be `(null if offsetExists is false)`

library/Requests/Utility/CaseInsensitiveDictionary.php
```diff
@@ -46,7 +46,7 @@ public function offsetExists($key) {
 	 * Get the value for the item
 	 *
 	 * @param string $key Item key
-	 * @return string Item value
+	 * @return string|null Item value (always a string if offsetExists is true)
```
rmccue: Again here, `(null if offsetExists is false)`

library/Requests/IRI.php
```diff
@@ -67,47 +67,47 @@ class Requests_IRI {
...
-	 * @var string
+	 * @var string|null
 	 */
 	protected $port = null;
 
 	/**
 	 * ipath
 	 *
-	 * @var string
+	 * @var string|null
```
jrfnl: Is this correct ? The default for the property is an empty string, not `null`.

library/Requests/IRI.php
```diff
@@ -118,6 +118,7 @@ class Requests_IRI {
 	 *
 	 * Each key is the scheme, each value is an array with each key as the IRI
 	 * part and value as the default value for that part.
+	 * @var array
```
jrfnl: Please add an empty comment line above the first tag.

jrfnl:
@TysonAndre I'd like to merge this in. Would you mind rebasing the PR so it becomes mergable ?
If you don't want to, don't know how or don't have the time, please let me know.

PR #271, opened 2017-03-24
Namespacify

DArcMattr:
I work on a WordPress site that a vendor built LiveFyre integration via its Composer-hosted library, which in turn has this Requests library as a dependency. I tracked down a weird fatal error which gets triggered by use of the native WordPress Requests class, where the code loads this library instead. 

I chose to rework the code to make use of namespacing. I made sure that all the example code & unit tests work as well as the original codebase, and along the way, I reworked the tests to use PHPUnit 6, which I included as a dev dependency. 

So here's my take on one of the things necessary to propel the PHP minimum requirement to 5.3, as mentioned in: https://github.com/rmccue/Requests/issues/138

This code has been battle tested in a production environment for a couple of weeks now.


rmccue:
Hi, and thanks for the PR.

Requests is used by WordPress, which requires PHP 5.2+. Until that requirement is changed, we cannot change. (There's also no great reason to.)

In addition, this would be a huge break of compatibility, and has minimal gain. To properly use namespaces would require redesigning the architecture of the classes as well (rather than just replacing `_` with `\`). I'd rather avoid breaking compatibility if at all possible. See #278 for some details on that.

Again, thanks, but closing out. If you have a specific issue caused by the bundling of Requests with WP, I'd be happy to consider fixing whatever that is.

fritzmg:
`Requests` is a pretty common name. Having the class in the root namespace is not a good idea. Should definitely be considered for `2.x`.

PR #280, opened 2017-06-05
Fix unit tests by aliasing PHPUnit_Framework_TestCase with namespaced name

### LINKED ISSUE #279: Fix unit tests
PHPUnit was upgraded on Travis, and this has broken all the tests, which use 5.2-style (`PHPUnit_Framework_TestCase`)

TysonAndre:
This change is intended to work in php 5.2+, the lowest version .travis.yml
indicates this supports.

If PHPUnit_Framework_TestCase does not exist, check if
PHPUnit\Framework\TestCase does exist, and if so, alias it.

- Different versions of phpunit may be required for different php
  versions, so don't try standardizing for now.

Fixes #279

### DIFF
--- a/tests/IRI.php
+++ b/tests/IRI.php
@@ -387,11 +387,11 @@ class RequestsTest_IRI extends PHPUnit_Framework_TestCase
 	/**
 	 * @expectedException PHPUnit_Framework_Error_Notice
 	 */
-	public function testNonexistantProperty()
+	public function testNonexistentProperty()
 	{
 		$iri = new Requests_IRI();
-		$this->assertFalse(isset($iri->nonexistant_prop));
-		$should_fail = $iri->nonexistant_prop;
+		$this->assertFalse(isset($iri->nonexistent_prop));
+		$should_fail = $iri->nonexistent_prop;
 	}
 
 	public function testBlankHost()
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ matrix:
   fast_finish: true
   include:
     - php: 5.2
+      dist: precise
     - php: 5.3
+      dist: precise
     - php: 5.4
     - php: 5.5
     - php: 5.6
@@ -22,6 +24,7 @@ cache:
 
 install:
  # Setup the test server
+ - phpenv install --releases
  - phpenv local 5.5
  - composer install --dev --no-interaction
  - TESTPHPBIN=$(phpenv which php)
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ matrix:
     - php: 5.6
       env: TEST_COVERAGE=1
     - php: 7.0
-    - php: hhvm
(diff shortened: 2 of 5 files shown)


TysonAndre:
https://travis-ci.org/rmccue/Requests/jobs/239492433 failed for the below reason

https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html

> You should be as specific as possible when testing exceptions. Testing for classes that are too generic might lead to undesirable side-effects. Accordingly, testing for the Exception class with @expectedException or setExpectedException() is no longer permitted.

Will fix this later, or rebase if another PR fixes that.

rmccue:
@TysonAndre The way I read that message is that when you call `setExpectedException`, you can't say you're expecting instances of `Exception`, but you can use more specific ones.

Based on https://github.com/sebastianbergmann/phpunit/issues/2074 it seems like we just need to use `$this->expectException` instead.

rmccue:
🙌 Finally passing!

Needs a review from a committer though :)

tests/IRI.php
```diff
@@ -387,11 +387,11 @@ public function testWriteAliased()
...
 	 * @expectedException PHPUnit_Framework_Error_Notice
 	 */
-	public function testNonexistantProperty()
+	public function testNonexistentProperty()
 	{
 		$iri = new Requests_IRI();
-		$this->assertFalse(isset($iri->nonexistant_prop));
-		$should_fail = $iri->nonexistant_prop;
+		$this->assertFalse(isset($iri->nonexistent_prop));
+		$should_fail = $iri->nonexistent_prop;
```
rmccue: (These changes aren't relevant, I just happened to notice the misspelling as I was fixing it.)

rmccue:
@TysonAndre Thanks for helping out with this issue (and on the others) by the way; you now have commit access. :)

ptasker:
I've tried to get the tests running a few times and it keeps bailing on the `fsockopen` Transport tests. Any tips on getting those running? 

PHPUnit 6.2.3
PHP 7.1.3
```
There were 3 failures:

1) RequestsTest_Transport_fsockopen::testResponseByteLimit
Failed asserting that 99 matches expected 104.

/Users/petertasker/working/Requests/tests/Transport/Base.php:48

2) RequestsTest_Transport_fsockopen::testStreamToFile
Failed asserting that null matches expected 'http://requests-php-tests.herokuapp.com/get'.

/Users/petertasker/working/Requests/tests/Transport/Base.php:489

3) RequestsTest_Transport_fsockopen::testMultipleToFile
Failed asserting that null matches expected 'http://requests-php-tests.herokuapp.com/get'.

/Users/petertasker/working/Requests/tests/Transport/Base.php:749
```

ozh:
Screw it. I'm done.

rmccue:
Thanks for this @TysonAndre, we ended up fixing it finally in #303.

PR #284, opened 2017-06-13
Fixes #283 : Fixes CPU busy loop when using request_multiple.

### LINKED ISSUE #283: performance issue when using request_multiple
The following code will cause serious performance problem:

```php
do {
    $status = curl_multi_exec($multihandle, $active);
}
while ($status === CURLM_CALL_MULTI_PERFORM);
```

it will be much better if changed as follow

```php
do {
    $status = curl_multi_exec($multihandle, $active);
    usleep(1000);
}
while ($status === CURLM_CALL_MULTI_PERFORM);
```

TysonAndre:
This will require more testing across various libcurl versions.
I doubt that the timeout is necessary for curl_multi_select
(calls select() if it can),
but leaving in one just in case of bugs, so that it will end.
- Haven't thoroughly checked for relevant libcurl bugs.

Asynchronously wait for events with a short timeout if CURLM_CALL_MULTI_PERFORM


library/Requests/Transport/cURL.php
```diff
@@ -219,8 +219,20 @@ public function request_multiple($requests, $options) {
...
+				// Timeouts can be specified in microseconds.
+				// Provide curl_multi_select with a timeout of 100ms (May not be necessary).
+				// As a fallback, if curl_multi_select couldn't properly asynchronously wait for input, sleep for 0.1 ms.
+				$select_status = curl_multi_select($multihandle, 0.1);
+				if ($select_status === -1) {
+					usleep(100);
+				}
 			}
-			while ($status === CURLM_CALL_MULTI_PERFORM);
+			while (true);
```
TysonAndre: Actually, the above approach may be incorrect. Maybe the part that was causing the busy loop wasn't the above `do{}while`, but below 

The below is probably a better alternative, and can get rid of the above.

```
                       while ($status === CURLM_CALL_MULTI_PERFORM);
 
+                       // Insert a 50ms sleep here so that we don't busy loop and chew cpu
+                       $selectRes  = curl_multi_select($multihandle, 0.05);
+
                        $to_process = array();
```

TysonAndre:
I'm encountering brand new issues due to https://github.com/php/php-src/pull/3509#issuecomment-430764448

On September 2018 for PHP 7.1.23, **after I created this PR,** curl_multi_select will now return `-1` without setting any file descriptors in certain error modes. That's because it uses curl_multi_wait(), which has slightly different issues.

I think that the issue might be triggered by timeouts (e.g. connect timeouts on one request, while the other request doesn't timeout).

https://bugs.php.net/bug.php?id=76480 mentions the following:

{quote}
Please note the patch changes the behaviour of curl_multi_select(). Previously curl_multi_select() would return -1 if cURL has no open file descriptors. In the new situation (using curl_multi_wait() with libcurl 7.28 or higher) curl_multi_select() will return 0 if cURL has no open file descriptors.
{quote}

One fix to consider would be to treat all remaining curl handles in the multi-handle as if they had timed out if curl_multi_select returned -1.

- It shouldn't be possible to return -1 in older php versions, because it looks like Requests properly checks if there are 1 or more remaining handles
- So if it does return -1, it's probably caused by a timeout, and should be handled by rmccue/Requests like that, throwing Timeout exceptions for the remaining connections

--------

Note that I'm not a curl_multi expert (I had a bit of experience using it from university, but that's it), and there may be edge cases I haven't considered

TysonAndre:
https://github.com/rmccue/Requests/pull/284#issuecomment-545495019 applied to an alternative version I was testing out locally without the usleep.

- The usleep is unnecessary if curl_multi_select properly slept for 50 milliseconds while waiting for an event.
- The usleep would mitigate CPU busy loop issues in newer php versions + curl versions. However, I think that request_multiple may still fail to return before/after my patch in edge cases, and adding `$Descriptors === -1 && curl_multi_errno( $Master ) !== CURLM_OK` may be needed

TysonAndre:
The failure mode I'm seeing in the latest php versions may be related to my use of the NOSIGNAL patch, if nobody else sees it

https://github.com/rmccue/Requests/issues/264#issuecomment-308214611

TysonAndre:
In order to reproduce the issue I saw, I did the following:

- Configured two requests: one to localhost which would connect immediately, and wait 3000 milliseconds to respond, and another to a service that would refuse to connect.
- Set a connect timeout of 3 milliseconds and a request timeout of 3000 milliseconds.
- Performed both requests

https://gist.github.com/TysonAndre/46a6829d9d8de00e50226890a6a9a63f has the full logs and details about the options used, but the part of interest is that curl_multi_select isn't sleeping for nearly as long as I'd expect it to, and it is neither returning -1 nor returning an error in curl_multi_error.

```
curl_multi_exec   returned 0 after 0.000492 seconds: curl_multi_errno=0
curl_multi_select returned 0 after 0.000010 seconds: curl_multi_errno=0
curl_multi_info_read returned a value:
/path/to/vendor/rmccue/requests/library/Requests/Transport/cURL.php:247:
array(3) {
  'msg' =>
  int(1)
  'result' =>
  int(7)
  'handle' =>
  resource(316) of type (curl)
}
curl error name: Couldn't connect to server
curl_multi_exec   returned 0 after 0.000013 seconds: curl_multi_errno=0
curl_multi_select returned 0 after 0.000005 seconds: curl_multi_errno=0
curl_multi_exec   returned 0 after 0.000081 seconds: curl_multi_errno=0
curl_multi_select returned 0 after 0.000025 seconds: curl_multi_errno=0
curl_multi_exec   returned 0 after 0.000031 seconds: curl_multi_errno=0
curl_multi_select returned 0 after 0.000002 seconds: curl_multi_errno=0
curl_multi_exec   returned 0 after 0.000031 seconds: curl_multi_errno=0
curl_multi_select returned 0 after 0.000004 seconds: curl_multi_errno=0
curl_multi_exec   returned 0 after 0.000008 seconds: curl_multi_errno=0
... 5000 more lines of busy looping
```

The below link has the code changes I added to diagnose the issue and the (ineffective) mitigations I'm currently trying. A hardcoded usleep() of 1 millisecond would harm PHP response times if the service would normally complete in less than a millisecond

```php
	/**
	 * Send multiple requests simultaneously
	 *
	 * @param array $requests Request data
	 * @param array $options Global options
	 * @return array Array of Requests_Response objects (may contain Requests_Exception or string responses as well)
	 */
	public function request_multiple($requests, $options) {
		// If you're not requesting, we can't get any responses ¯\_(ツ)_/¯
		if (empty($requests)) {
			return array();
		}

		$multihandle = curl_multi_init();
		$subrequests = array();
		$subhandles = array();

		$class = get_class($this);
		foreach ($requests as $id => $request) {
			$subrequests[$id] = new $class();
			$subhandles[$id] = $subrequests[$id]->get_subrequest_handle($request['url'], $request['headers'], $request['data'], $request['options']);
			$request['options']['hooks']->dispatch('curl.before_multi_add', array(&$subhandles[$id]));
			curl_multi_add_handle($multihandle, $subhandles[$id]);
		}

		$completed = 0;
		$responses = array();

		$request['options']['hooks']->dispatch('curl.before_multi_exec', array(&$multihandle));

		do {
			$active = false;

			$is_first_multi_exec = true;
			do {
				/*
				if (!$is_first_multi_exec) {
					// Sleep for 1 millisecond to avoid a busy loop that would chew CPU.
					// XXX uncomment this to mitigate this a bit - there are better mitigations such as making this sleep after curl_multi_select if it returns 0 or -1.
					usleep(1000);
				}
				 */
				$start = microtime(true);
				$status = curl_multi_exec($multihandle, $active);
				fprintf(STDERR, "curl_multi_exec   returned %s after %.6f seconds: curl_multi_errno=%s\n", json_encode($status), microtime(true) - $start, curl_multi_errno($multihandle));
				$is_first_multi_exec = false;
			}
			while ($status === CURLM_CALL_MULTI_PERFORM);

			// Insert a 50ms sleep here so that we don't busy loop and chew cpu
			$start = microtime(true);
			$selectRes  = curl_multi_select($multihandle, 2.0);
			fprintf(STDERR, "curl_multi_select returned %s after %.6f seconds: curl_multi_errno=%s\n", json_encode($selectRes), microtime(true) - $start, curl_multi_errno($multihandle));
			fflush(STDERR);

			$to_process = array();

			// Read the information as needed
			while ($done = curl_multi_info_read($multihandle)) {
				echo "curl_multi_info_read returned a value:\n";
				var_dump($done);
				echo "curl error name: " . curl_strerror($done['result']) . "\n";
				$key = array_search($done['handle'], $subhandles, true);
				if (!isset($to_process[$key])) {
					$to_process[$key] = $done;
				}
			}

			// Parse the finished requests before we start getting the new ones
			foreach ($to_process as $key => $done) {
				$options = $requests[$key]['options'];
				if (CURLE_OK !== $done['result']) {
					//get error string for handle.
					$reason = curl_error($done['handle']);
					$exception = new Requests_Exception_Transport_cURL(
									$reason,
									Requests_Exception_Transport_cURL::EASY,
									$done['handle'],
									$done['result']
								);
					$responses[$key] = $exception;
					$options['hooks']->dispatch('transport.internal.parse_error', array(&$responses[$key], $requests[$key]));
				}
				else {
					$responses[$key] = $subrequests[$key]->process_response($subrequests[$key]->response_data, $options);

					$options['hooks']->dispatch('transport.internal.parse_response', array(&$responses[$key], $requests[$key]));
				}

				curl_multi_remove_handle($multihandle, $done['handle']);
				curl_close($done['handle']);

				if (!is_string($responses[$key])) {
					$options['hooks']->dispatch('multiple.request.complete', array(&$responses[$key], $key));
				}
				$completed++;
			}
		}
		while ($active || $completed < count($subrequests));

		$request['options']['hooks']->dispatch('curl.after_multi_exec', array(&$multihandle));

		curl_multi_close($multihandle);

		return $responses;
	}
```
</section>

TysonAndre:
It seems like the addition of usleep doesn't hurt short requests, but makes long requests take noticeably longer. I should only be performing the usleep if the number returned by curl_multi_select wasn't positive.

https://curl.haxx.se/libcurl/c/curl_multi_wait.html

> On completion, if numfds is non-NULL, it will be populated with the total number of file descriptors on which interesting events occurred. This number can include both libcurl internal descriptors as well as descriptors provided in extra_fds.

The implementation in php-src for curl_multi_select:

```c
#if LIBCURL_VERSION_NUM >= 0x071c00 /* Available since 7.28.0 */
	error = curl_multi_wait(mh->multi, NULL, 0, (unsigned long) timeout * 1000.0, &numfds);
	if (CURLM_OK != error) {
		SAVE_CURLM_ERROR(mh, error);
		RETURN_LONG(-1);
	}

	RETURN_LONG(numfds);
#else
	_make_timeval_struct(&to, timeout);

	FD_ZERO(&readfds);
	FD_ZERO(&writefds);
	FD_ZERO(&exceptfds);

	error = curl_multi_fdset(mh->multi, &readfds, &writefds, &exceptfds, &maxfd);
	SAVE_CURLM_ERROR(mh, error);

	if (maxfd == -1) {
		RETURN_LONG(-1);
	}
	RETURN_LONG(select(maxfd + 1, &readfds, &writefds, &exceptfds, &to));
#endif
```

TysonAndre:
No further changes are planned. This seems to be working pretty well to mitigate edge cases causing cpu usage spikes in a fraction of requests.

- After the latest commit, requestMultiple duration is back to what it was without the sleep for normal requests (durations ranging from 5-1000ms)
- Overall CPU usage is slightly down on hosts that are using requestMultiple a lot (with php >= 7.1.23 and curl >= 7.28.0). CPU usage stops spiking when the request duration of the external service spikes
- The measured number of calls to curl_multi_exec is a lot lower.

jrfnl:
@TysonAndre Hiya! Sorry for the long wait, but we're finally trying to get through the list of open PRs.

Also: would you be willing and able to add some tests for these changes to the PR ?

jrfnl:
PR #493 is possibly a partial duplicate of this PR.

Issue #295, opened 2017-10-20
Test failure with recent PHPUnit version


remicollet:
Discovered in Fedora QA, since 5.7.23
https://apps.fedoraproject.org/koschei/package/php-rmccue-requests?collection=f28

Also see https://github.com/zendframework/zend-mail/pull/174

@sebastianbergmann another one broken by the empty message change

remicollet:
```
There were 122 failures:
1) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #15 (400, false)
Failed asserting that exception message is empty but is '400 Bad Request'.
2) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #16 (401, false)
Failed asserting that exception message is empty but is '401 Unauthorized'.
3) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #17 (402, false)
Failed asserting that exception message is empty but is '402 Payment Required'.
4) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #18 (403, false)
Failed asserting that exception message is empty but is '403 Forbidden'.
5) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #19 (404, false)
Failed asserting that exception message is empty but is '404 Not Found'.
6) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #20 (405, false)
Failed asserting that exception message is empty but is '405 Method Not Allowed'.
7) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #21 (406, false)
Failed asserting that exception message is empty but is '406 Not Acceptable'.
8) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #22 (407, false)
Failed asserting that exception message is empty but is '407 Proxy Authentication Required'.
9) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #23 (408, false)
Failed asserting that exception message is empty but is '408 Request Timeout'.
10) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #24 (409, false)
Failed asserting that exception message is empty but is '409 Conflict'.
11) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #25 (410, false)
Failed asserting that exception message is empty but is '410 Gone'.
12) RequestsTest_Transport_cURL::testStatusCodeThrow with data set #26 (411, false)
Failed asserting that exception message is empty but is '411 Length Required'.
[...]

PR #297, opened 2017-10-23
fix for recent change in PHPUnit

remicollet:
See issue #295


ozh:
I think you're connecting unrelated issues. See #298. Feel free to comment on if you think I'm mistaken.

PR #298, opened 2017-10-23
Fix the tests

ozh:
This PR "fixes" the tests. Emphasis on "fixes", because some tests fail, but the whole process isn't interrupted by some unwanted PHP error

**Summary of changes:**
- add PHPUnit in composer for tests, and run this version of PHPUnit instead of the one bundled with the PHP build. This makes sure the next change works:
- replace `PHPUnit_Framework_TestCase` with `PHPUnit\Framework\TestCase` and use PHPUnit's backward compat feature
- Specify Travis distro for use

**Known problems, need more work**
- I removed the early `phpenv` call which was breaking all the build except on PHP 5.5 ([example](https://travis-ci.org/ozh/Requests/jobs/291654326#L463)). No idea why. As a consequence I think it won't be possible to run tests on 5.2 which, despite WordPress's paleolithic requirement, isn't a big deal I think.
- For some goddam unknown reason, the test server breaks on PHP 5.4, seems to be an autoload problem but I haven't found wtf.

All in al,l now some tests are failing but the process as a whole is saner (except for 5.4 that is), I think that's a better coding base to work on failing tests again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
 		"php": ">=5.2"
 	},
 	"require-dev": {
+        "phpunit/phpunit": ">=4.0",
 		"requests/test-server": "dev-master"
 	},
 	"type": "library",
--- a/tests/Transport/cURL.php
+++ b/tests/Transport/cURL.php
@@ -1,5 +1,5 @@
 <?php
 
-class RequestsTest_Transport_cURL extends RequestsTest_Transport_Base {
+class RequestsTest_Transport_cURL extends PHPUnit\Framework\TestCase {
 	protected $transport = 'Requests_Transport_cURL';
 }
--- a/tests/Transport/cURL.php
+++ b/tests/Transport/cURL.php
@@ -1,5 +1,5 @@
 <?php
 
-class RequestsTest_Transport_cURL extends PHPUnit\Framework\TestCase {
+class RequestsTest_Transport_cURL extends RequestsTest_Transport_Base {
 	protected $transport = 'Requests_Transport_cURL';
 }
--- a/tests/Transport/fsockopen.php
+++ b/tests/Transport/fsockopen.php
@@ -1,5 +1,5 @@
 <?php
 
-class RequestsTest_Transport_fsockopen extends PHPUnit\Framework\TestCase {
+class RequestsTest_Transport_fsockopen extends RequestsTest_Transport_Base {
 	protected $transport = 'Requests_Transport_fsockopen';
 }
--- a/tests/Transport/fsockopen.php
+++ b/tests/Transport/fsockopen.php
@@ -1,5 +1,5 @@
 <?php
 
-class RequestsTest_Transport_fsockopen extends RequestsTest_Transport_Base {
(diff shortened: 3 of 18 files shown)


ozh:
Note: I think this supersedes #280 since no class aliasing is required

composer.json
```diff
@@ -14,6 +14,7 @@
 		"php": ">=5.2"
 	},
 	"require-dev": {
+        "phpunit/phpunit": ">=4.0",
```
remicollet: Need to check the minimal version providing the "forword compat" stuff

AFAIK: ^4.8.35 || ^5.7 || ^6.0

remicollet:
I can't even understand how this can work... Transport/Base use setExpectedException which have been removed in PHPUnit 6...

tests/Transport/cURL.php
```diff
@@ -1,5 +1,5 @@
 <?php
 
-class RequestsTest_Transport_cURL extends RequestsTest_Transport_Base {
+class RequestsTest_Transport_cURL extends PHPUnit\Framework\TestCase {
```
remicollet: No => RequestsTest_Transport_Base else, no test

tests/Transport/fsockopen.php
```diff
@@ -1,5 +1,5 @@
 <?php
 
-class RequestsTest_Transport_fsockopen extends RequestsTest_Transport_Base {
+class RequestsTest_Transport_fsockopen extends PHPUnit\Framework\TestCase {
 	protected $transport = 'Requests_Transport_fsockopen';
```
remicollet: No => RequestsTest_Transport_Base else, no test

remicollet:
So, if you fix curl.php and fsockopen.php tu extends the Base test suite

With phpunit 5 => 
`Failed asserting that exception message is empty but is '400 Bad Request'.`

With phpunit6 =>
`Error: Call to undefined method RequestsTest_Transport_fsockopen::setExpectedException()`

remicollet:
So with PHPUnit 5.7 and 6.x you have to use expectException and expectExceptionCode (as in PR #297), but these don't exist in 4.8, so need setExpectedException there...

composer.json
ozh: Indeed, 4.8.35, but this is not needeed since composer checks out the latest version compatible with the current PHP version it is testing and 4.8.35 works on PHP 5.3+ (yes, this will screw PHP 5.2)

tests/Transport/cURL.php
ozh: Indeed, good catch

remicollet:
So, if you look at travis result you will observe failure describe in my previous comment

So please test with #299

ozh:
Holy crap some jobs are now passing 😄

ozh:
At this point, 5.3, 5.4 and possibly 5.2 need the "phpenv local 5.5" trick before calling composer and starting the local server, but [this fails](https://travis-ci.org/rmccue/Requests/builds/245481972) and I don't know why. Ping @rmccue.

rmccue:
Handled in #303 instead, thanks for all your work on this @ozh, and thanks for the feedback @remicollet.

PR #299, opened 2017-10-24
Issue phpunit

remicollet:
Refresh of #297 for #295 
Tested with PHPUnit 4.8.35 and 5.7.23

#298 also needed for PHPUnit 6


ozh:
Indeed. I'm integrating this in the PR being worked on, with less duplicated code.

PR #302, opened 2017-12-12
Fix Travis CI

ntwb:
This is a WIP to fix Travis CI - ~Do Not Merge Yet~ 🤖 

Cherry picking various pieces from #298 and #280

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,7 @@ sudo: false
 cache:
   directories:
     - $HOME/.cache/pip
-    - $HOME/.composer/cache
-    - vendor
+    - $HOME/.composer/cache/files
 
 install:
  # Setup the test server
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,10 @@ cache:
 
 install:
  # Setup the test server
- - phpenv local 5.5
+ - phpenv global 5.5
  - composer install --dev --no-interaction
  - TESTPHPBIN=$(phpenv which php)
- - phpenv local --unset
+ - phpenv global --unset
 
  # Setup the proxy
  - pip install --user mitmproxy~=0.15
@@ -60,7 +60,7 @@ script:
 after_script:
  - utils/proxy/stop.sh
  - cd ..
- - phpenv local 5.5
+ - phpenv global 5.5
  - PATH=$PATH vendor/bin/stop.sh
  - test $TEST_COVERAGE && bash <(curl -s https://codecov.io/bash)
- - phpenv local --unset
+ - phpenv global --unset
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ before_script:
  - curl -s -I http://requests-php-tests.herokuapp.com/ > /dev/null
  
  # Environment checks
- ../vendor/bin/phpunit --version
+ - ../vendor/bin/phpunit --version
 
 script:
  - ../vendor/bin/phpunit --coverage-clover clover.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,24 @@
+dist: trusty
 language: php
 matrix:
   fast_finish: true
   include:
     - php: 5.2
+      dist: precise
     - php: 5.3
+      dist: precise
     - php: 5.4
     - php: 5.5
     - php: 5.6
       env: TEST_COVERAGE=1
     - php: 7.0
+    - php: 7.1
+    - php: 7.2
+    - php: nightly
     - php: hhvm
+
+allow_failures:
+  - php: nightly
 
 # Use new container infrastructure
 sudo: false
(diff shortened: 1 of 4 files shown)


ntwb:
This "could" be merged now, at least the PHPUnit failures are now visible, fixing those should get passing Travis CI jobs for PHP 5.3 - PHP 7.2

I haven't even looked at any HHVM issues, I expect them to fail and should more than likely be removed, though I'm not sure who else apart from WordPress is using this library

PHP 5.2 is broken, because Composer requires >PHP 5.3, I can follow up fixes for PHP 5.2 in a subsequent PR.

ntwb:
If I can can get someone to approve this I'll merge it 🎉

.travis.yml
```diff
@@ -1,31 +1,37 @@
...
-    - $HOME/.composer/cache
-    - vendor
+    - $HOME/.composer/cache/files
 
 install:
  # Setup the test server
- - phpenv local 5.5
  - composer install --dev --no-interaction
  - TESTPHPBIN=$(phpenv which php)
- - phpenv local --unset
```
rmccue: Does `composer install` work on PHP 5.2 though? That was the issue previously which required the `phpenv` hack.

ntwb: It does not, I'll add back Composer for PHP 5.2 in a follow-up PR, it was quicker and easier to have some CI functionality for the majority of PHP versions rather than nothing, now all but PHP 5.2 and HHVM work, previously all CI jobs were erroring

rmccue:
I'm hesitant to merge a PR that doesn't actually fix the tests. Based on the failures I see on https://travis-ci.org/rmccue/Requests/jobs/315173786, it seems like the local HTTP server isn't actually running, which I suspect is due to removing the `phpenv` lines.

ntwb:
That’s fair enough, I hadn’t dug in any further than confirming PHPUnit tests were now running (previously they were not)

I’m also pretty unfamiliar with this library, I’ll take a look at what the HTTP server bits look like and take a stab at fixing that, and PHP 5.2

As to fixing the bugs reported I’ll be leaving that others cc @dd32

dd32:
Looks like the proxy failure is due to mitmproxy not supporting Python 2:

```
Error: mitmproxy requires Python 3.5 or above
Starting with version 1.0 released in 12/2016, mitmproxy no longer supports Python 2.
To install the latest version of mitmproxy using pip on Python 3, run:
    pip uninstall mitmproxy
    pip3 install mitmproxy
You can also use our standalone binaries that come with their own Python interpreter:
    http://docs.mitmproxy.org/en/stable/install.html
To get rid of this message and use the last version of mitmproxy that supports Python 2, run:
    pip install "mitmproxy==0.18.2"
Apologies for the inconvenience!`

**Comment by dd32** [comment]
I took a further swing at it, here's the green result: https://travis-ci.org/dd32/Requests/builds/315356440 from https://github.com/dd32/Requests/tree/pr/302 specifically https://github.com/dd32/Requests/commit/3bbcd2bd55b0f29a42ebb6c886201f24e10b6a1e, https://github.com/dd32/Requests/commit/80291d262501acbe87b36284c74409e88e9e39e4, https://github.com/dd32/Requests/commit/4b91d503f025adde368d1cd6b48476660f22c56a, and https://github.com/dd32/Requests/commit/6493d638d901a9ecc06bc9715e5f2fd2cd6e4944 (I can't push to the PR here)

- MITMProxy was failing as 1.0+ requires Python3, seemed easier to just pin it to the last release which supports Python2. This also required fixing the proxy.py script, I couldn't see what version the syntax of the callbacks changed.
- `phpenv local` was failing on `precise` as it seems that `phpenv` doesn't have the X.Y -> X.Y.Z symlinks created for PHP 5.5/5.6 (only for 5.2/5.4). Instead i dynamically pulled the version in (5.6.13/5.6.31 on precise/trusty)
- HHVM failed because of silly PHPUnit version compares, I just dropped the HHVM testing entirely https://github.com/dd32/Requests/commit/4b91d503f025adde368d1cd6b48476660f22c56a (see https://travis-ci.org/dd32/Requests/jobs/315315609) If you want to keep it @rmccue it's possible, but doesn't seen worth anyones time.
- Tests fail occasionally on the following test, but that's unrelated to the issue at hand, restarting the job fixes that.
```
RequestsTest_Transport_cURL::testMultipleWithFailure
Undefined property: Requests_Exception_Transport_cURL::$status_code
/home/travis/build/dd32/Requests/tests/Transport/Base.php:682
```

ozh:
@dd32 you rock. I totally gave up on phpenv when trying to fix. I completely agree on HHVM.

So, what's next ? We close here and you open a new PR to merge it?

rmccue:
@dd32 Thanks for that! Let's PR that up and get it merged.

> HHVM failed because of silly PHPUnit version compares

Happy to drop compatibility with it now that they're no longer aiming for PHP compatibility.

Issue #306, opened 2018-02-10
Requests_Cookie::parse not setting host-only


soulseekah:
I was writing tests for https://core.trac.wordpress.org/ticket/43231 and stumbled across this unexpected behavior:

```php
$cookie = Requests_Cookie::parse( 'test=foo; domain=.wordpress.org' );
$cookie->flags['host-only'] == false // assertion error!
```

The `host-only` flag is not being set, although it should be, shouldn't it?

Thoughts on enforcing the host-flag here: https://github.com/rmccue/Requests/blob/master/library/Requests/Cookie.php#L290

> // Domain normalization, as per RFC 6265 section 5.2.3

Which states that the host-only flag should be set.

rmccue:
The `host-only` flag represents the

[Per §5.3.2](https://tools.ietf.org/html/rfc6265#section-5.2.3):
```
   If the first character of the attribute-value string is %x2E ("."):

      Let cookie-domain be the attribute-value without the leading %x2E
      (".") character.

   Otherwise:

      Let cookie-domain be the entire attribute-value.
```

The `host-only` flag is set during the processing, following the rules from [§5.3](https://tools.ietf.org/html/rfc6265#section-5.3):
```
   6.   If the domain-attribute is non-empty:

           If the canonicalized request-host does not domain-match the
           domain-attribute:

              Ignore the cookie entirely and abort these steps.

           Otherwise:

              Set the cookie's host-only-flag to false.

              Set the cookie's domain to the domain-attribute.

        Otherwise:

           Set the cookie's host-only-flag to true.

           Set the cookie's domain to the canonicalized request-host.
```

This is done in the [parse_from_headers](https://github.com/rmccue/Requests/blob/4055bc412f128bb298735d5e1b0189eff41f1340/library/Requests/Cookie.php#L445-L451) method.

Not seeing that there's a bug here?

PR #307, opened 2018-02-10
Enforce host-only flag in `normalize_attribute`

soulseekah:
When a non host-only flag domain is supplied, during its normalization
(as per RFC 6265 section 5.2.3) not only should the dot be removed but
the host-only flag should be set to false.

This fixes weird behavior in `Requests_Cookie::parse()`

See #306

Tests included.

### DIFF
--- a/library/Requests/Cookie.php
+++ b/library/Requests/Cookie.php
@@ -291,6 +291,7 @@ class Requests_Cookie {
 				// Domain normalization, as per RFC 6265 section 5.2.3
 				if ($value[0] === '.') {
 					$value = substr($value, 1);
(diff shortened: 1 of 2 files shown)


soulseekah:
Fixed the `testDomainExactMatch` and `testDomainMatch` tests. Since this patch enforces the host-only flag, exact matching in some test cases will fail. Overriding the flag after cookie creation (and normalization) allows for exact matching in the tests.

soulseekah:
This infringes on the RFC, as to set the host-only flag to false we need to know the request-host (Host header). Closing as invalid. Sorry.

PR #309, opened 2018-02-10
Add support for various deflate compression levels

### LINKED ISSUE #301: Requests::decompress falsely determines content is not compressed
https://github.com/rmccue/Requests/blob/master/library/Requests.php#L823

That function attempts to poke at the data to see if it is actually compressed or not. On a Ubuntu 16.04 system without php-curl installed the fsockopen transport is used instead of curl. When accessing content that is return using encoding deflate or gzip (what I tested) this routine falsely determines that the content is NOT compressed and returns the compressed data leading to failures in dependent software.

My environment consists of the following:

* WordPress 4.9.1
* ElasticPress 2.4
* Ubuntu 16.04 with PHP 7.0 (without php-curl installed)
* ElasticSearch 5.2.2
* WP-CLI 1.4

Once configured, using `wp elasticpress stats` will report that it was unable to contact the ElasticSearch server. This is due to the data coming back compressed (using deflate directly out of ElasticSearch) but being determined as not compressed despite actually being compressed. Not performing the check or installing php-curl on the system will resolve the issue.

soulseekah:
Different compression levels yield a specific second byte in the magic
header for the deflate encoding. All of them can be decoded by the same
functions without any issues. Let them through, as they've been seen in
the wild.

Fixes #301

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -821,7 +821,9 @@ class Requests {
 	 * @return string Decompressed string
 	 */
 	public static function decompress($data) {
-		if (substr($data, 0, 2) !== "\x1f\x8b" && substr($data, 0, 2) !== "\x78\x9c") {
+		// All valid deflate, gzip header magic markers
+		$valid_magic = array("\x1f\x8b", "\x78\x01", "\x78\x5e", "\x78\x9c", "\x78\xda");
+		if (!in_array(substr($data, 0, 2), $valid_magic)) {
 			// Not actually compressed. Probably cURL ruining this for us.
 			return $data;
 		}


library/Requests.php
```diff
@@ -821,7 +821,9 @@ public static function flattern($array) {
 	 * @return string Decompressed string
 	 */
 	public static function decompress($data) {
-		if (substr($data, 0, 2) !== "\x1f\x8b" && substr($data, 0, 2) !== "\x78\x9c") {
+		// All valid deflate, gzip header magic markers
+		$valid_magic = array("\x1f\x8b", "\x78\x01", "\x78\x5e", "\x78\x9c", "\x78\xda");
+		if (!in_array(substr($data, 0, 2), $valid_magic)) {
```
rmccue: This needs to use the `$strict` parameter to `in_array`, but otherwise looks good.

rmccue (dismissed):
This also needs tests in the [Encoding.php](https://github.com/rmccue/Requests/blob/master/tests/Encoding.php) suite.

soulseekah:
Added strict in_array checking, added compression level tests (generated using zpipe.c from the zlib 1.2.11 using levels 1, 3 and 9, 5 was already being tested as the generic case). Also added Encoding.php to the test suite, it was not explicitly defined in phpunit.xml.dist and thus was not being run at all.

library/Requests.php
soulseekah: Fixed in follow up commit. Tests included.

jrfnl:
@schlessera I've just been looking at this PR. This seems like a good change. What do you think ?

To get this ready to merge, IMO, the following actions are needed:
* Rebase the PR.
    This should automatically remove the change to `tests/phpunit.xml.dist`. If not, that change needs to be removed as this is now fixed more thoroughly already.
* Move the `$valid_magic` array to a class constant and document the source of the magic markers.
    Nice to have/micro-optimization: Use the markers as the keys in the array instead of as the values and use `isset()` instead of `in_array()`.

jrfnl:
I've made the changes as per the above comment. This PR is now ready for review/merge.

src/Requests.php
```diff
@@ -184,6 +184,25 @@ class Requests {
 	 */
 	protected static $certificate_path;
 
+	/**
+	 * All (known) valid deflate, gzip header magic markers.
+	 *
+	 * These markers related to different compression levels.
```
schlessera:
```suggestion
	 * These markers relate to different compression levels.
```

src/Requests.php
```diff
@@ -187,7 +187,7 @@ class Requests {
 	/**
 	 * All (known) valid deflate, gzip header magic markers.
 	 *
-	 * These markers related to different compression levels.
+	 * These markers relate to different compression levels.
```
soulseekah, line 190: Thanks :)

schlessera:
Thanks for the PR, @soulseekah !

Issue #310, opened 2018-02-11
verify => false does not work with fsockopen


soulseekah:
The `Requests_Transport_fsockopen` still throws an exception regardless of the `verify` option on invalid SSL. Test case incoming.

PR #311, opened 2018-02-11
verify => false does not work with fsockopen

soulseekah:
See #310 fsockopen will fail


Gerk:
Did this ever get pulled/addressed?  I've run up against this issue and really don't want to have to move to another library as this one suits my needs very nicely (aside from this issue that is)

jrfnl:
@soulseekah Thank you for this PR and sorry for the long wait.

This looks like a good change. Would you mind rebasing the PR so it becomes mergable ?
If you don't want to, don't know how or don't have the time, please let me know.

soulseekah:
Rebased. Thank you.

schlessera (approved):

PR #316, opened 2018-02-17
Add tests for 303 status code

soulseekah:
Tests for issue #7

### DIFF
--- a/tests/Transport/BaseTestCase.php
+++ b/tests/Transport/BaseTestCase.php
@@ -1182,6 +1182,6 @@ abstract class BaseTestCase extends TestCase {
 		$request = Requests::post(httpbin('/status/303', true), array(), $data, $this->getOptions(array('follow_redirects' => true)));
 
 		$this->assertSame(200, $request->status_code);
-		$this->assertSame(httpbin('/get'), $request->url);
+		$this->assertSame('/get', substr($request->url, -4));
 	}
 }
--- a/tests/Transport/BaseTestCase.php
+++ b/tests/Transport/BaseTestCase.php
@@ -1173,11 +1173,6 @@ abstract class BaseTestCase extends TestCase {
 	}
 
 	public function test303GETmethod() {
-		if ($this->skip_https) {
-			$this->markTestSkipped('SSL support is not available.');
-			return;
-		}
-
 		$data    = ['test' => 'true', 'test2' => 'test'];
 		$request = Requests::post(httpbin('/status/303', true), [], $data, $this->getOptions(['follow_redirects' => true]));
 
--- a/tests/Transport/BaseTestCase.php
+++ b/tests/Transport/BaseTestCase.php
@@ -1179,9 +1179,9 @@ abstract class BaseTestCase extends TestCase {
 		}
 
 		$data    = array('test' => 'true', 'test2' => 'test');
-		$request = Requests::post('https://httpbin.org/status/303', array(), $data, $this->getOptions(array('follow_redirects' => true)));
+		$request = Requests::post(httpbin('/status/303', true), array(), $data, $this->getOptions(array('follow_redirects' => true)));
 
 		$this->assertSame(200, $request->status_code);
-		$this->assertSame('https://httpbin.org/get', $request->url);
+		$this->assertSame(httpbin('/get'), $request->url);
 	}
 }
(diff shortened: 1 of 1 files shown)


soulseekah:
Done.

> Regarding the code, I only wonder whether assertEquals() is the right assertion to use. Should this possibly be assertSame() instead ?

Wouldn't matter too much, assertEquals is used throughout the tests both here and in WordPress and has not been an issue ever. Hope this helps. Thank you! :)

jrfnl:
> Wouldn't matter too much, assertEquals is used throughout the tests both here and in WordPress and has not been an issue ever. Hope this helps. Thank you! :)

I have a PR ready to, at the very least, change it for `Requests` (waiting for some other PRs to be merged first) and believe you me, it really *is* an issue as you're not actually testing what you intend to test and that can lead to horrible bugs (and horrible discussions ... "_it's tested and the tests pass, so not our bug_" / "_no, there really is a bug in your code_").

tests/Transport/Base.php
```diff
@@ -858,4 +858,12 @@ public function testBodyDataFormat() {
 		$this->assertEquals(httpbin('/post'), $result['url']);
 		$this->assertEquals(array('test' => 'true', 'test2' => 'test'), $result['form']);
 	}
+
+	public function test303GETmethod() {
+		$data = array('test' => 'true', 'test2' => 'test');
```
jrfnl: Looks like this tests needs a skip check like this to prevent the build failing on PHP 5.2.

https://github.com/rmccue/Requests/blob/347016983d0938e8bc4f76924dde3163cbbfa301/tests/Transport/Base.php#L518-L521

jrfnl:
I've updated the PR to address the previously made remark. @schlessera If you agree as well, this can be merged.

jrfnl (approved):

jrfnl:
Ok, so I'm clearly doing something wrong as the test output keeps changing and failing. Will have to revisit this.

tests/Transport/Base.php
```diff
@@ -894,4 +894,17 @@ public function testBodyDataFormat() {
...
+		if ($this->skip_https) {
+			$this->markTestSkipped('SSL support is not available.');
+			return;
+		}
+
+		$data    = array('test' => 'true', 'test2' => 'test');
+		$request = Requests::post(httpbin('/status/303', true), array(), $data, $this->getOptions(array('follow_redirects' => true)));
+
+		$this->assertSame(200, $request->status_code);
+		$this->assertsame(httpbin('/get'), $request->url);
```
schlessera, line 908:
```suggestion
		$this->assertSame(httpbin('/get'), $request->url);
```

schlessera (changes requested):

tests/Transport/Base.php
jrfnl: Fixed, but as function names are case-insensitive, it's not the reason the test keeps failing.

tests/Transport/Base.php
```diff
@@ -894,4 +894,17 @@ public function testBodyDataFormat() {
...
+		if ($this->skip_https) {
+			$this->markTestSkipped('SSL support is not available.');
+			return;
+		}
+
+		$data    = array('test' => 'true', 'test2' => 'test');
+		$request = Requests::post(httpbin('/status/303', true), array(), $data, $this->getOptions(array('follow_redirects' => true)));
+
+		$this->assertSame(200, $request->status_code);
+		$this->assertSame(httpbin('/get'), $request->url);
```
schlessera, line 908: Comparing `httpbin('/get')` (the response of a request) to `request->url` (a URL string) does not make sense here.

The check should assert that the method that was used for `REQUEST` was `GET`, so how about this:
```suggestion
		$this->assertSame('/get', substr($request->url, -4));
```

schlessera (changes requested):

jrfnl (approved):
Confirmed that the test fails without the fix from https://github.com/WordPress/Requests/commit/969bfdfdc27519226c1ed9b471b4097ab3b0205a#diff-71e2db25b8f80f81de28ddb29b507afaa10518556b9b7997cf97bdeb3dd532e1R413-R415 and passes with it. Let's (finally) get this merged!

schlessera (approved):

PR #318, opened 2018-02-17
Force Content-Length for POST requests

soulseekah:
Addresses #248


jrfnl:
Duplicate of #249

jrfnl:
I've created PR #368 which consolidates both this as well as PR #249. Feedback welcome.

Issue #257, opened 2016-12-08
Failed to open stream: Too many open files in Requests.php


DennisGoldfingerTM:
While using `Requests` in a script that shoots out multiple thousands of requests, I get the following error:

```
PHP Warning:  require_once(/.../Requests/Exception.php): failed to open stream: Too many open files in /.../Requests.php on line 136
PHP Fatal error:  require_once(): Failed opening required '/.../Requests/Exception.php' (include_path='.:/usr/local/lib/php:/var/www/...') in /.../Requests.php on line 136
```

I believe this indicates that file handles are not being closed properly. It also seems to drive up memory usage linearly (probably a result of open file handles).

soulseekah:
Are you able to recollect what the actual request parameters were? Method, options, sample of posted data, etc.?

DennisGoldfingerTM:
It has been a while since I reported this issue, but I was able to recall the context. The best I can give you is the class method used in the script I mentioned. Here is the full method (note that there is also a GET variant that is very similar):

```
/**
 * A function that shoots out a POST request to X API
 *
 * @param $endpoint
 * @param $params
 * @param bool | array $debug - false when turned off, array of options when on
 * @return array
 */
private function _post($endpoint, $params, $debug = false) {
    $return = array('success' => false);
    $response = null;
    $attempts = 5;
    do {
        try {
            $response = Requests::post($this->_url($endpoint), ['Content-Type' => 'application/json'], json_encode($params), ['transport' => 'Requests_Transport_fsockopen']);
        } catch(Exception $e) {
            $attempts--;
            sleep(1);
        }
    } while(empty($response) && $attempts > 0);
    
    if(!empty($response) && $response->success) {
        if($debug) {$this->_dump($response->body, $debug['dumpAsJson']); exit;}
        $body = json_decode($response->body, true);
        if($body['success']) {
            $return['success'] = true;
            $return['data'] = $body['data'];
        } else {
            $return['errorMessage'] = $body['errorMessage'];
        }
    }
    if($debug) {$this->_dump($response); exit;}
    return $return;
}
```

The context here is that we are using this class to query an external API through a regularly run cron script. The script, at the time, would query this API thousands of times over the course of 6-12 hours (probably around 5000 requests total). The long run time wasn't due to the requests themselves but the processing that occurred after each request, so it wasn't a timeout related issue.

The script has run smoothly ever since the **transport** parameter was set to **Requests_Transport_fsockopen** as I mentioned in my original post.

[...]

Issue #39, opened 2013-02-01
Memory leak in cURL transport


PawelDecowski:
I discovered a memory leak in the cURL transport. When doing multiple `Requests::request()` calls (for example in a loop) the memory usage keeps growing (in my case by 3–5MB every call).

I’ve narrowed it down to `$this->fp` not being closed in the cURL transport’s `request()` method. Now, this shouldn’t cause the memory leak, as a transport is instantiated in the scope of the `Requests::request()` method so it (and thus all its instance variables, including `$this->fp`) should be released when that method call completes.

I tried looking for a reference to the transport which is not released but can’t find any.

The temporary solution is to add `curl_close($this->fp)` just before the `Requests_Transport_cURL::request()` returns. Although it does fix the problem, it does not fix it at the source.

Before I fork your repo and add the temporary fix I thought I’d let you know so maybe you can find the real cause of this bug — since you’re more familiar with your code.

PR #319, opened 2018-02-17
Prevent cURL transport from leaking on Exception

soulseekah:
The `CURLOPT_HEADERFUNCTION` and `CURLOPT_WRITEFUNCTION` options are not
being reset if a cURL handle error occurs (which throws a
`Requests_Exception` and stops execution). The destructor is not called
due to lingering instance method callbacks in the two options.

Move `CURLOPT_HEADERFUNCTION` and `CURLOPT_WRITEFUNCTION` cleanup before
any exceptions can happen.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -174,12 +174,14 @@ class Requests_Transport_cURL implements Requests_Transport {
 			$response = $this->response_data;
 		}
 
-		$this->process_response($response, $options);
+		if (true === $options['blocking']) {
+			// Need to remove the $this reference from the curl handle.
+			// Otherwise Requests_Transport_cURL wont be garbage collected and the curl_close() will never be called.
+			curl_setopt($this->handle, CURLOPT_HEADERFUNCTION, null);
+			curl_setopt($this->handle, CURLOPT_WRITEFUNCTION, null);
+		}
 
-		// Need to remove the $this reference from the curl handle.
-		// Otherwise Requests_Transport_cURL wont be garbage collected and the curl_close() will never be called.
-		curl_setopt($this->handle, CURLOPT_HEADERFUNCTION, null);
-		curl_setopt($this->handle, CURLOPT_WRITEFUNCTION, null);
+		$this->process_response($response, $options);
 
 		return $this->headers;
 	}


soulseekah:
Calling `curl_close` has been chased time (#39) and time (#59) again. Found another one while investigating #257 

Requesting a bad URL, even if it's a mere certificate error, will not call the destructor and eventually `curl_close`. Why?

`Requests_Transport_cURL::process_response` can throw an Exception. This happens BEFORE the `CURLOPT_HEADERFUNCTION` and `CURLOPT_WRITEFUNCTION` are freed of the instance callbacks, so the garbage collector will not collect `$this` resulting in an out of memory error.

Test case:

```php
while(true) {                                                                                                                                                                                  
    try {
        $response = Requests::post('http://localhost:43992', array(), array(), array('transport'=>'Requests_Transport_cURL'));
    } catch (Exception $e ) {}
}
```

Where http://localhost:43992 is a nonexistent local endpoint (to error out faster).

Calling `curl_close` has been chased time (#39) and time (#59) again. Found another one while investigating #257 

Requesting a bad URL, even if it's a mere certificate error, will not call the destructor and eventually `curl_close`. Why?

`Requests_Transport_cURL::process_response` can throw an Exception. This happens BEFORE the `CURLOPT_HEADERFUNCTION` and `CURLOPT_WRITEFUNCTION` are freed of the instance callbacks, so the garbage collector will not collect `$this` resulting in an out of memory error.

Breaking case:

```php
while(true) {                                                                                                                                                                                  
    try {
        $response = Requests::post('http://localhost:43992', array(), array(), array('transport'=>'Requests_Transport_cURL'));
    } catch (Exception $e ) {}
}
```

Where http://localhost:43992 is a nonexistent local endpoint (to error out faster).

gorlovka:
Leaks exception for me too:
http://i.kagda.ru/780166178436_03-06-2020-23:52:00_7801.png

Trying to catch is unsuccessful

silviucpp:
There is one big problem with the fix proposed here: `curl_setopt` called before `process_response` will reset the `curl_errno` (used in `process_response`) state. And the library will trigger an exception with "Missing header/body separator" instead real cause " Connection refused". 

A better fix will be:

```php
try
{
    $this->process_response($response, $options);
}
finally
{
    if (true === $options['blocking'])
    {
        // Need to remove the $this reference from the curl handle.
        // Otherwise Requests_Transport_cURL wont be garbage collected and the curl_close() will never be called.
        curl_setopt($this->handle, CURLOPT_HEADERFUNCTION, null);
        curl_setopt($this->handle, CURLOPT_WRITEFUNCTION, null);
    }
}
```

schlessera:
As we have now moved to PHP 5.6 as a minimum requirement, it seems to make sense to take a fresh look at the execution flow, and use the `finally` keyword to close the cURL handle safely no matter what.

jrfnl:
Note for the next iteration of this PR:

For testing the correct functioning of this change, the PHPUnit 9.3+ `Assert::assertIsClosedResource()` assertion can be used. This assertion is [polyfilled by the PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills/#phpunit--930-yoastphpunitpolyfillspolyfillsassertclosedresource), so can be safely used.

Also note: tests are only relevant for PHP < 8.0 and should be skipped on PHP 8.0+ as [Curl was changed from using resources to using objects](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.resource2object).

schlessera:
Tests are failing because PHP 7 does check for active references to a resource before closing it, even if you explicitly use `curl_close()`. In that case, it does not actually close the resource as requested.

In our case, the very fact that we want to test whether the resource is closed properly creates a reference to the resource. So while there is no memory leak in the normal code path, there is one in the test - because of the test.

The only way I can think of for solving this is by using weak references, but those have only been introduced with PHP 7.2.

Options right now:
- Don't test for the closed resource.
- Use a weak reference to test for the closed resource, but skip these tests on PHP 7.0 and PHP 7.1.

jrfnl:
Okay, so I have spend some time today trying to create a better/working test for this. Current status can be seen here: https://github.com/WordPress/Requests/tree/319/better-test-attempt-take-two

Tentative conclusions:
* The current fix doesn't solve the problem. The `Curl` object instance still doesn't get destroyed.
    It may well be a step in the right direction, but it doesn't fix the issue _yet_.
* There seems to be a reference somewhere to the `$options` array (containing a reference to the $transport class) which will need to be unset, which we haven't been able to trace back yet. This needs further investigation to figure out where the `$options` arrays gets stored.

PR #327, opened 2018-08-03
 post request support multi-file upload

jianminLee:
Just add a $files file path array parameter, the file array can be a two-dimensional array, upload file can also support complex array request parameter request.

example：
``` php 
Requests::post($url,$header,['test' => ['a' => 'b','c' => [1,2,3 => ['cc' => ['vv' => [456,789,321]]]]]],[
        'paid_certificates' => [
            public_path('images/not_found.jpg'),
            public_path('images/abg.jpg'),
            public_path('images/bg.jpg'),
        ],
    ]);
```

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -265,35 +265,35 @@ class Requests {
 	/**
 	 * Send a POST request/Support for multiple files, complex array parameters
 	 */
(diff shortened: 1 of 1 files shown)


soulseekah:
Related: https://github.com/rmccue/Requests/pull/313

Let's make sure we have the same API, please.

PR #328, opened 2018-08-28
Add a .gitattributes file

KasperFranz:
This is added to keep the file size when using composer down

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 bin/ export-ignore
 docs/ export-ignore
-example/ export-ignore
+examples/ export-ignore
 tests/ export-ignore
 .coveralls.yml export-ignore
 .gitignore export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+bin/ export-ignore
+docs/ export-ignore
+example/ export-ignore
+tests/ export-ignore
+.coveralls.yml export-ignore
+.gitignore export-ignore
+.gitattributes export-ignore
+.travis.yml export-ignore
+package.xml.tpl export-ignore


KasperFranz:
@rmccue  any chance this can get implemented?

rmccue:
@KasperFranz Apologies for the delay here. Happy to merge in; could you merge master into your branch so tests pass please?

rmccue:
@KasperFranz It appears you've rebased here, as the commits now appear as committed by you but authored by someone else. Could you correct this please?

KasperFranz:
It was a quick test, while not being at my machine at home

I will do that later today

rmccue:
No rush, this repo moves fairly slowly :)

ozh:
I'm in favor of ignoring those. I would even remove all non php files, in particular all the .MD files that are much less readable on a server than on Github. The idea is that when using the library as is on a prod server, you don't end up with files that can be used for [full path disclosure](https://www.owasp.org/index.php/Full_Path_Disclosure)

KasperFranz:
> 
> 
> Only thing I'm wondering about is the `ignore` for the `docs` and the `examples`.
> 
> All the rest isn't needed for someone new to the repo to implement it, but those can be useful for a first time implementation (similar to the readme).

My thoughts were that when you are using this (or any other package), I would typically look through the code and the documentation online instead of reading through a docs folder in the vendor directory.

This was partly made to start a discussion about it as well, I am glad to make changes to the code if anyone wants me to change anything.

The branch is also up to date now.

jrfnl:
> I would even remove all non php files, in particular all the .MD files that are much less readable on a server than on Github.

The way the `.gitattributes` file is now in this PR, it basically does, safe for the `cacert.pem` file, the readme, changelog and license file and I would never recommend for those to be `export-ignore`d.

PR #334, opened 2019-03-01
Update stats on PHP 5.2 usage as of March 2019

tnorthcutt:


ozh:
@rmccue tests failing because feelingrestful.com is down, any other humanmade option?

jrfnl:
@tnorthcutt I'd like to merge this in. Would you mind updating it with the latest % (1.5%) and rebasing the PR so it becomes mergable ?

If you don't want to or don't have the time, please let me know.

tnorthcutt:
@jrfnl I would love to, but I both don't have time right now, and I think I'm not clear on the numbers at this point. WP Stats shows 1.5% on PHP 5.2, but that doesn't really make sense if the number was recently 60%...

I'd prefer to leave this to someone else. Thanks for the ping.

jrfnl:
> I would love to, but I both don't have time right now, and I think I'm not clear on the numbers at this point. WP Stats shows 1.5% on PHP 5.2, but that doesn't really make sense if the number was recently 60%...

If I remember correctly, the colors for PHP 5.2 and PHP 5.6 at some point were the same and caused confusion. PHP 5.6 at the time was at 60%. (seen this issue elsewhere as well).

jrfnl:
New PR with props created - see #367

PR #336, opened 2019-04-04
Support HTTP/2

tijszwinkels:


rmccue:
Hi, sorry for not getting back to you earlier. HTTP/2 support requires some significant changes to the Requests transports internally, as it's a vastly different protocol at the socket layer (which is where the sockets transport operates). As such, closing this, as it doesn't add that functionality.

Thanks anyway!

PR #337, opened 2019-04-18
max_bytes not respected in a performant manner

dd32:
With some testing I was noticing that a request for 100 byes from a 10MB file was timing out after 5 seconds.

Turns out that the cause was that Requests downloads the full 10MB file in order to return the first 100 bytes of it.

This PR simply aborts the connection when enough bytes have been read from the stream.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -166,12 +166,12 @@ class Requests_Transport_cURL implements Requests_Transport {
 
 		$curl_errno = curl_errno($this->handle);
 
-		if ( $curl_errno === 23 && $this->response_byte_limit && $this->response_byte_limit === $this->response_bytes ) {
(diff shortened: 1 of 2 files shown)


dd32:
Requests Unit tests are broken, however the unit test which previously covered this functionality should still continue to pass.

ozh:
Damn, I thought approving the PR would be enough to merge but no, tests must pass too, and they're all broken because feelingrestful.com is dead :(

rmccue:
This is my bad! cc @joehoyle any chance you can take a look at that?

rmccue:
@dd32 Could you please merge master into this branch so we can look at merging? ✨

schlessera:
This PR was closed inadvertently when we changed the stable branch from `master` to `stable`. GitHub was supposed to smoothly adapt the PRs to point to the new branch, but this seems to have failed for the cases where the author's branch/fork of the PR is not available anymore.

jrfnl:
Recreated as #494

PR #339, opened 2019-04-28
Ignore locale when creating the HTTP version string from a float

### LINKED ISSUE #335: Always get "Bad request (400)"
I just got started with Requests, using it with php5.6.

For some reason I get the response "Bad request" (400) on every GET and POST request I try.

I've tested this:
   $response = Requests::get('https://github.com/timeline.json');

with a number of URLs and they do respond (I get a github page) but they all say "Bad request" (400).

Tried with POST as well, the example in the docs:
			$url = 'https://api.github.com/some/endpoint';
			$headers = array('Content-Type' => 'application/json');
			$data = array('some' => 'data');
			$response = Requests::post($url, $headers, json_encode($data));
			var_dump($response->status_code);
But it also gives 400.

Sorry for the noob question, but where do I begin to find the cause for this?
Is it because I'm using php 5.6?

Thanks!

Zegnat:
This should fix #335.

Rather than using `%f`, this should use the non-locale aware `%F`. That way PHP will always use a period (`.`) for the decimal separator in the HTTP version, and never fallback to a comma (`,`).

I wasn’t sure how to add a test for this, as I’d want to intercept the socket message somehow. But to illustrate the issue:

```
php > echo setlocale(LC_NUMERIC, 0);
C
php > echo sprintf("%s %s HTTP/%.1f\r\n", 'GET', '/', 1.1);
GET / HTTP/1.1
php > echo setlocale(LC_NUMERIC, 'de_DE');
de_DE
php > echo sprintf("%s %s HTTP/%.1f\r\n", 'GET', '/', 1.1);
GET / HTTP/1,1
php > echo sprintf("%s %s HTTP/%.1F\r\n", 'GET', '/', 1.1);
GET / HTTP/1.1
```

### DIFF
--- a/library/Requests/Transport/fsockopen.php
+++ b/library/Requests/Transport/fsockopen.php
@@ -150,7 +150,7 @@ class Requests_Transport_fsockopen implements Requests_Transport {
 		$options['hooks']->dispatch('fsockopen.remote_host_path', array(&$path, $url));
 
 		$request_body = '';
-		$out = sprintf("%s %s HTTP/%.1f\r\n", $options['type'], $path, $options['protocol_version']);
+		$out = sprintf("%s %s HTTP/%.1F\r\n", $options['type'], $path, $options['protocol_version']);
 
 		if ($options['type'] !== Requests::TRACE) {
 			if (is_array($data)) {


Zegnat:
Not sure why #335 was closed as a support issue while the bug identified in there, and patched here, is still there.

Remerged with master as per https://github.com/rmccue/Requests/pull/346#issuecomment-531606632 in the hope to get all tests to pass and get this PR accepted. Something still seems wrong with Travis? Seeing errors connecting to localhost, even from cURL when this patch only touches the fsocket code.

Let me know what I can do to get this resolved!

rmccue:
Reopened #335, apologies.

The tests are likely failing due to transient issues; the nature of working with servers is that occasionally things fail for no reason. I've restarted those tests.

rmccue:
Tests are now passing after restarting, thanks!

Zegnat:
Many thanks for merging! 🎊

jrfnl:
I've opened PR #521 to add a regression test for this fix.

PR #345, opened 2019-06-28
Fix unit tests

datagutten:
The tests currently use an empty string or the default value for checking exception message which means empty message. The message is not empty and causes the tests to fail. I changed this to null so the message is not checked and those tests pass.


rmccue:
Thanks so much for this pull request and apologies for the delay. I've rolled these tests into #351 to also update the domain name; will aim to merge that.

PR #351, opened 2019-09-15
Fix unit tests by updating domain and PHPUnit versions

rmccue:
feelingrestful.com was decommissioned.

This should 🤞 fix the tests and allow other PRs to be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,8 @@ matrix:
      env: COMPOSER_PHPUNIT=true
    - php: nightly
      env: COMPOSER_PHPUNIT=true
-
-allow_failures:
-  - php: nightly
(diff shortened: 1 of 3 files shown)


rmccue:
This should pass now, as all tests are passing except `nightly` (although it was incorrectly ignored previously). I've incorporated @datagutten's changes from #345 to fix the PHPUnit classes (thanks!).

Once that's done, @jrfnl's #346 should fix the failures on "nightly" (we're actually missing 7.3 as well; this will likely fix that too).

rmccue:
Tests are now passing; I had to restart a couple due to the usual transient failures from external servers.

rmccue:
Thanks @ntwb!

PR #352, opened 2019-09-16
Travis: test against PHP 7.3 and 7.4

jrfnl:
As `Request` is supposed to work cross-version, testing against PHP 7.3 (out since dec 2018) and PHP 7.4 (expected end of november 2019) would be a good idea.


jrfnl:
> I assume we'll need to change 7.4snapshot to 7.4 upon release?

You should be able to update it as soon as Travis adds an official PHP 7.4 image. They normally do this once a PHP version goes into RC, but my latest tests to see if it was available yet have been unsuccessful, so waiting patiently...

PR #353, opened 2019-09-16
Travis: cache the correct Composer directories

jrfnl:


jrfnl:
FYI: The PHP 7.0 build needs to be restarted (unrelated failure).

PR #355, opened 2019-09-16
Unit tests: attempt to make the build more stable

jrfnl:
The `testAlternatePort` tests for both Curl as well as fsockopen often time out and cause builds to fail.

This PR attempts to make the builds more stable, by:
1. Retrying the request a second time if the first request timed out.
2. Skipping the unit test if the second request timed out as well.

Adding `verbose=true` to the PHPUnit config will make sure that if the tests are skipped, the skip reason will always be displayed when running the tests.

### DIFF
--- a/tests/Transport/Base.php
+++ b/tests/Transport/Base.php
@@ -765,7 +765,20 @@ abstract class RequestsTest_Transport_Base extends PHPUnit_Framework_TestCase {
 	}
 
 	public function testAlternatePort() {
-		$request = Requests::get('http://portquiz.net:8080/', array(), $this->getOptions());
+		try {
+			$request = Requests::get('http://portquiz.net:8080/', array(), $this->getOptions());
+		} catch( Requests_Exception $e ) {
+			// Retry the request as it often times-out.
+			try {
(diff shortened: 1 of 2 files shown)


jrfnl:
> Just to verify, this won't affect whether those pass/fail (although will print more verbose information about them)?

Correct, all this does is list the tests which were skipped and, if available, the reason why they were skipped at the bottom of the test output.

jrfnl:
FYI: just added a second commit applying the same fix to the tests in the `RequestsTest_Proxy_HTTP` class.

rmccue:
The proxy shouldn't fail, as it's a locally running instance of mitmproxy, so I'd rather not skip those tests. Rather than retrying, we may just need to increase the timeout to account for the additional time needed to connect to the proxy.

jrfnl:
> The proxy shouldn't fail, as it's a locally running instance of mitmproxy, so I'd rather not skip those tests. Rather than retrying, we may just need to increase the timeout to account for the additional time needed to connect to the proxy.

I'll remove that commit in that case. I just looked at it as the earlier build failure for #353 was related to that.

PR #357, opened 2019-09-17
PHP compatibility: don't use short arrays

jrfnl:
Short arrays were introduced in PHP 5.4.
As this package is supposed to be compatible with PHP 5.2 and up, they cannot be used and IMO, even the example code should comply with that.

### DIFF
--- a/examples/cookie_jar.php
+++ b/examples/cookie_jar.php
@@ -7,13 +7,13 @@ include('../library/Requests.php');
 Requests::register_autoloader();
 
 // Say you need to fake a login cookie
-$c = new Requests_Cookie_Jar(['login_uid' =>  'something']);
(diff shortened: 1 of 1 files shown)


rmccue:
I'm torn on this one. I see the point, but I also think we should encourage using more modern solutions by default. (Also, I'd like to drop PHP <5.6 at least in the next major release.)

jrfnl:
> I'm torn on this one. I see the point, but I also think we should encourage using more modern solutions by default.

In that case, I'll close this PR and will add a switch in the upcoming PR which adds linting to the Travis script as otherwise, the builds would not pass.

>  (Also, I'd like to drop PHP <5.6 at least in the next major release.)

:+1:

PR #360, opened 2019-09-19
CS: variety of whitespace fixes

jrfnl:
_[This PR is part of a series to introduce code style checking to the repo]_

Fix whitespace issues:
* Don't use space indentation.
* Don't use inline tabs.
* Don't use precision alignment
* No multiple consecutive blank lines.
* One space (or a new line) on either side of a concatenation operator.
* Always have a space (or new line) before and after each operator, with the exception of the boolean not `!` and the spread operator.
* No space between the boolean not `!` and what it applies to.
    Note: _This is the opposite of WPCS, but in line with the majority of uses of the `!` operator in this repo._
* Always have a space (or new line) after a comma.
* Don't use horizontal alignment within single line arrays.
* One space between the close parenthesis and the opening curly of function declarations.
* No space on the inside of function call, function declaration, array, control structure or arbitrary parentheses.
    Note: _This is the opposite of WPCS, but in line with the majority of code in this repo._
    Also note: for arrays this won't be enforced as there is no sniff available for this, but the current code-base has been fixed where necessary.
* Multi-line function calls should have each parameter on a new line.
* There should be one space between a control structure keyword and its opening parenthesis.
* For multi-part control structures - `if/else`, `try/catch`, the next keyword should be on a new line.
    Note: _This is the opposite of WPCS, but in line with the majority of code in this repo._
    Also note: This will not be enforced as there is no sniff for it, but is in line with the prevailing code style in the repo.
* End each file with a new line character.


jrfnl:
Rebased for merge conflicts.

schlessera (approved):

jrfnl:
Rebased for merge conflicts ;-)

PR #368, opened 2019-09-26
Add Content-* headers for empty POST requests

### LINKED ISSUE #248: `Content-Length` header not specified for empty POST requests with fsockopen transport

jrfnl:
This PR consolidates the work done by @dd32 in PR #249 and @soulseekah in PR #318 which both addressed the same issue.

Fixes #248
Closes #249
Closes #318

### DIFF
--- a/library/Requests/Transport/fsockopen.php
+++ b/library/Requests/Transport/fsockopen.php
@@ -162,6 +162,8 @@ class Requests_Transport_fsockopen implements Requests_Transport {
 				$request_body = $data;
 			}
 
+			// Always include Content-length on POST requests to prevent
+			// 411 errors from some servers when the body is empty.
 			if (!empty($data) || $options['type'] === Requests::POST) {
 				if (!isset($case_insensitive_headers['Content-Length'])) {
 					$headers['Content-Length'] = strlen($request_body);
@@ -169,12 +171,6 @@ class Requests_Transport_fsockopen implements Requests_Transport {
 
 				if (!isset($case_insensitive_headers['Content-Type'])) {
 					$headers['Content-Type'] = 'application/x-www-form-urlencoded; charset=UTF-8';
-				}
-			}
-			elseif($options['type'] === Requests::POST) {
-				if (!isset($case_insensitive_headers['Content-Length'])) {
-					// Prevent 411 errors from some servers
-					$headers['Content-Length'] = 0;
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


dd32:
As the author of #249 this looks good to me, not having actually run the code in several years.

gstrauss:
The HTTP method POST by definition contains an entity body.  As such, the Content-Length request header must be set, or the entity body must be sent via HTTP/1.1 Transfer-Encoding: chunked (or HTTP/2).

There are numerous prior tickets (#249, #318) and plenty of blog posts about this issue.  Additionally, this bug *reintroduced* a bug fixed 12 years ago in Wordpress.  https://core.trac.wordpress.org/ticket/8605

The patch in this PR is simple and looks reasonable, and this PR is over a year old.  #249 is from 2016 (!)

What is holding up actioning this PR, committing and releasing a fix?

library/Requests/Transport/fsockopen.php
```diff
@@ -160,8 +160,9 @@ public function request($url, $headers = array(), $data = array(), $options = ar
 				$request_body = $data;
 			}
 
-			if (!empty($data)) {
+			if (!empty($data) || $options['type'] === Requests::POST) {
 				if (!isset($case_insensitive_headers['Content-Length'])) {
+					// Prevent 411 errors from some servers when the body is empty.
```
schlessera, line 165: The comment is in the wrong spot, as it actually explains the extra condition, not the header string.
```suggestion
			// Always include Content-length on POST requests to prevent
			// 411 errors from some servers when the body is empty.
			if (!empty($data) || $options['type'] === Requests::POST) {
				if (!isset($case_insensitive_headers['Content-Length'])) {
```

schlessera (changes requested):

library/Requests/Transport/fsockopen.php
jrfnl: Fixed up

schlessera (approved):

gstrauss:
Thank you for picking this up, @schlessera, and thank you, @jrfnl, for jumping back in to quickly incorporate feedback to make the patch ready to merge.

jrfnl:
@gstrauss Case of striking the iron when it's hot and such ... :grinning:

PR #371, opened 2019-10-05
Add support for sending POST with multipart/form-data

datagutten:
Some sites require POST requests to be sent as multipart/form-data.
When passing an array as POST-data cURL defaults to multipart/form-data, but Requests will flatten and urlencode it. To avoid this I check if a Content-Type header containing multipart/form-data is set in the arguments to Requests::post.
I have implemented this for both cURL and fsockopen.


jrfnl:
@datagutten Thank you for contributing. Would it be possible to add some unit tests for this change ?

PR #375, opened 2019-12-11
Feature to set ip resolver type in request options

ShivamYuvraj:


jrfnl:
@ShivamYuvraj Thank you for contributing. Would it be possible to add some unit tests for this change ?

ozh:
Any idea on how to implement a unit test that would not add yet another dependency on an external host?

PR #376, opened 2020-01-30
Updates cacert.pem file to a more recent version

jitendra-1217:
__Changelog__
- [x]  Updates cacert.pem file to a more recent version from [this source](https://curl.haxx.se/docs/caextract.html).

  There has been 2 more open prs but still filled new as there has been more updates to cacert bundle since then.

  Alternatively, long term fix would be to not have cacert bundle committed in code. But this would require more though and change. Request to merge this.


jitendra-1217:
Closing this as more recent pr, with an updated bundle, for the same are open.

PR #379, opened 2020-03-15
Throw exception if invalid filename is provided

datagutten:
Throw an exception if an invalid path is provided in the filename option.
Before this change, the script will continue and try to write and close a file pointer which does not exist.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -156,7 +156,11 @@ class Requests_Transport_cURL implements Requests_Transport {
 		$options['hooks']->dispatch('curl.before_send', array(&$this->handle));
 
 		if ($options['filename'] !== false) {
-			$this->stream_handle = fopen($options['filename'], 'wb');
+			@$this->stream_handle = fopen($options['filename'], 'wb');
+			if($this->stream_handle===false) {
+				$error = error_get_last();
+				throw new Requests_Exception($error['message'], 'fopen');
+			}
 		}
 
 		$this->response_data       = '';
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -156,8 +156,9 @@ class Requests_Transport_cURL implements Requests_Transport {
 		$options['hooks']->dispatch('curl.before_send', array(&$this->handle));
 
 		if ($options['filename'] !== false) {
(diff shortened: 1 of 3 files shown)


jrfnl:
@datagutten  Hiya! Sorry for the long wait, but we're finally trying to get through the list of open PRs.

library/Requests/Transport/cURL.php
```diff
@@ -135,7 +135,11 @@ public function request($url, $headers = array(), $data = array(), $options = ar
 		$options['hooks']->dispatch('curl.before_send', array(&$this->handle));
 
 		if ($options['filename'] !== false) {
-			$this->stream_handle = fopen($options['filename'], 'wb');
+			@$this->stream_handle = fopen($options['filename'], 'wb');
```
schlessera: The error silencing should be as limited as possible, so we want ot have it directly in front of the problematic `fopen()` call.
```suggestion
			$this->stream_handle = @fopen($options['filename'], 'wb');
```

library/Requests/Transport/fsockopen.php
```diff
@@ -234,7 +234,11 @@ public function request($url, $headers = array(), $data = array(), $options = ar
 		$doingbody = false;
 		$download = false;
 		if ($options['filename']) {
-			$download = fopen($options['filename'], 'wb');
+			@$download = fopen($options['filename'], 'wb');
```
schlessera:
```suggestion
			$download = @fopen($options['filename'], 'wb');
```

schlessera (changes requested):

library/Requests/Transport/fsockopen.php
```diff
@@ -234,7 +234,11 @@ public function request($url, $headers = array(), $data = array(), $options = ar
...
+			if($download===false) {
```
schlessera:
```suggestion
			if($download === false) {
```

schlessera (changes requested):

jrfnl:
* Rebased the PR to make it mergable.
* Added an extra commit with minor tweaks after review.

jrfnl:
Added a second commit with minor tweaks to the test and adjustment to an existing test to allow for the exception. This should now be good to go.

schlessera (approved):

PR #387, opened 2020-06-18
Use more clear and inclusive language.

desrosj:
This changes "blacklisted" to "blocklisted". It is more clear and more inclusive language.


desrosj:
The failure looks to be a cURL connection issue.

ozh:
[ here I sat for 2 mn pondering if I should say something or not ]

PR #406, opened 2020-10-21
CS: move test helper classes to their own file

jrfnl:
### CS: move test helper classes to their own file [1]

* Add the new class file to a new `tests/Mocks` directory.
* Prefix the class, same as the other classes in the test directory.

### CS: move test helper classes to their own file [2]

* Add the new class file to the `tests/Mocks` directory.
* Prefix the class, same as the other classes in the test directory.


schlessera (changes requested):
Approach looks good, however the folder name and class name should use the singular `Mock` to be more correct, otherwise it would give the impression they represent a collection of mocks, instead of an individual mock.

jrfnl:
> Approach looks good, however the folder name and class name should use the singular Mock to be more correct, otherwise it would give the impression they represent a collection of mocks, instead of an individual mock.

Happy enough to change it, but that, to me, sounds similar to saying `Requests` should be called `Request` as it handles on request at a time. Or saying the `tests` directory should be called `test`. Or am I missing something ?

jrfnl:
Either way, the PR has been updated now.

schlessera:
`Requests` is the project name, and `tests` is a convention for that type of folders.

For classes, it is a long-standing and established convention to use the singular form for objects, except for collections. This is also deeply ingrained in most APIs and tools, part of the UML definition, etc...

I have a hard time finding a single authoritative source for this (which would probably be something rather old), but if you Google about whether to use singular or plural class names, you'll usually find the same answer consistently.

The rule of thumb is that all objects use the singular form apart from collections of objects. So you instantiate a `Book` object, not a `Books` object. However, you'd collect a multiple of `Book` objects in a `Books` collection.

schlessera (approved):

PR #409, opened 2020-10-22
CS: use `self` where appropriate

jrfnl:
Use self when referencing a (static) class member from within the same class.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -520,7 +520,7 @@ class Requests {
 			'idn'              => true,
 			'hooks'            => null,
 			'transport'        => null,
-			'verify'           => Requests::get_certificate_path(),
+			'verify'           => self::get_certificate_path(),
 			'verifyname'       => true,
 		);
 		if ($multirequest !== false) {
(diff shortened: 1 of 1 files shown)


ozh:
Is this good practice ? I mean, if at some point things are refactored and bits are moved to a subclass, it's easy to forget this.

What's the benefit ?

jrfnl:
@ozh There are up and down sides to this.
By the same logic you use above, the class could be renamed and this would be one use of the name which now won't need to be changed anymore because of using `self`.

Generally speaking, it increases the code readability as a dev now knows instantly that the method/property is intended to be one of *this* class.

schlessera:
Using `self`/`static` is considered good practice, indeed.

I'm going to merge this as is, but in case we raise the PHP minimum version to 5.3 or higher, we absolutely need to replace `self` with `static` (late static binding was only introduced with PHP 5.3), as this constitutes a bug. The problem here is that the class is not `final`, and right now if you'd extend the class to override `get_certificate_path()`, it would just break. a non-final class should in general just use `self` (or its class name, which is the same thing) when referring to a private property/method.

schlessera (approved):

jrfnl:
Agreed. I actually considered using `static` (forgetting that the repo has a 5.2 minimum) and for now decided against it as it would be a BC-break, considering that it originally used the hard-coded class name.

When older PHP support is dropped in the next major, such a BC-break would not be an issue.

PR #411, opened 2020-10-23
CS: don't use assignments in conditions

jrfnl:
Assignments in conditions makes for hard to debug, hard to read code and is considered a code smell.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -637,7 +637,8 @@ class Requests {
 		$return->url = $url;
 
 		if (!$options['filename']) {
-			if (($pos = strpos($headers, "\r\n\r\n")) === false) {
+			$pos = strpos($headers, "\r\n\r\n");
+			if ($pos === false) {
 				// Crap!
 				throw new Requests_Exception('Missing header/body separator', 'requests.no_crlf_separator');
 			}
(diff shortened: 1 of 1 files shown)


library/Requests.php
```diff
@@ -826,17 +827,30 @@ public static function decompress($data) {
 			return $data;
 		}
 
-		if (function_exists('gzdecode') && ($decoded = @gzdecode($data)) !== false) {
-			return $decoded;
+		if (function_exists('gzdecode')) {
+			$decoded = @gzdecode($data);
+			if ($decoded !== false) {
```
schlessera, line 832: This code not exactly the same. In the original code, the `&&` operator will cause the result of `$decoded = @gzdecode($data)` to be cast as a `bool` before being compared to `false`.

In the changed code, this casting doesn't happen, and considering we do a strict comparison, it would for example have a different result when `$decoded = @gzdecode($data)` would return `null`.

Now, according to the docs, `gzdecode()` would only ever return `false` on failure. But I'm unsure what it would return if you successfully decode a document of zero length. Would it return `null` then?

library/Requests.php
```diff
@@ -826,17 +827,30 @@ public static function decompress($data) {
...
+				return $decoded;
+			}
 		}
-		elseif (function_exists('gzinflate') && ($decoded = @gzinflate($data)) !== false) {
-			return $decoded;
+
+		if (function_exists('gzinflate')) {
+			$decoded = @gzinflate($data);
+			if ($decoded !== false) {
```
schlessera, line 839: This code is changing as well because of the stripped cast to `bool`.

library/Requests.php
```diff
@@ -826,17 +827,30 @@ public static function decompress($data) {
...
+		$decoded = self::compatible_gzinflate($data);
+		if ($decoded !== false) {
 			return $decoded;
 		}
-		elseif (function_exists('gzuncompress') && ($decoded = @gzuncompress($data)) !== false) {
-			return $decoded;
+
+		if (function_exists('gzuncompress')) {
+			$decoded = @gzuncompress($data);
+			if ($decoded !== false) {
```
schlessera, line 851: This code is changing as well because of the stripped cast to `bool`.

schlessera (commented):
The code is not a 100% match to its original version. I'm unsure about the implication, though.

Thoughts, @jrfnl ?

library/Requests.php
schlessera: I think the new version is actually more correct, but it would be a BC.

jrfnl: > Now, according to the docs, gzdecode() would only ever return false on failure. But I'm unsure what it would return if you successfully decode a document of zero length. Would it return null then?

Basing myself also on the documentation, I would expect it to return an empty string in that case as the standard return type is a string.

library/Requests.php
jrfnl: Same thing.

library/Requests.php
jrfnl: > This code not exactly the same. In the original code, the && operator will cause the result of $decoded = @gzdecode($data) to be cast as a bool before being compared to false.

Not sure if that would ever make a difference, so I'm not convinced this is a BC-break.
To proof my point: https://3v4l.org/OAoLN

schlessera (approved):

PR #414, opened 2020-10-24
QA: use the most specific PHPUnit assertion possible

jrfnl:
This is a long established best practice.

PHPUnit contains a variety of assertions and the ones available has been extended hugely over the years.

To have the most reliable tests, the most specific assertion should be used.

Refs:
* https://phpunit.de/manual/4.8/en/appendixes.assertions.html
* https://phpunit.readthedocs.io/en/7.5/assertions.html#

Most notably, this changes nearly all calls to `assertEquals()` to `assertSame()`, where `assertEquals()` does a loose type comparison and `assertSame()` does a strict type comparison.

Refs:
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertEquals
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertSame

---- 

Note: this change now causes two tests to fail on PHP 5.4-5.6 and three tests to fail on PHP 5.2-5.3.

Based on these test results, the methods under test are not PHP cross-version compatible as they return different results depending on the PHP version.

We'll need to decide what the desired behaviour should be and will need to adjust the code base to fix this. Input for this very welcome.

---
<details>
  <summary><b>Test failure details</b></summary>

The first two fail on all PHP 5.x builds. The third only on PHP 5.2 and 5.3.

```
There were 3 failures:
1) RequestsTest_Transport_cURL::testHEAD
Failed asserting that false is identical to ''.
/home/travis/build/rmccue/Requests/tests/Transport/Base.php:154
/home/travis/.phpenv/versions/5.2.17/bin/phpunit:46
2) RequestsTest_Transport_fsockopen::testHEAD
Failed asserting that false is identical to ''.
/home/travis/build/rmccue/Requests/tests/Transport/Base.php:154
/home/travis/.phpenv/versions/5.2.17/bin/phpunit:46
3) RequestsTest_Session::testURLResolution
Failed asserting that Requests_IRI Object (
    'scheme' => 'http'
    'iuserinfo' => null
    'ihost' => 'localhost'
    'port' => 8080
    'ipath' => '/get'
    'iquery' => null
    'ifragment' => null
    'normalization' => Array (
        'acap' => Array (
            'port' => 674
        )
        'dict' => Array (
            'port' => 2628
        )
        'file' => Array (
            'ihost' => 'localhost'
        )
        'http' => Array (
            'port' => 80
        )
        'https' => Array (
            'port' => 443
        )
    )
) is identical to 'http://localhost:8080/get'.
/home/travis/build/rmccue/Requests/tests/Session.php:10
/home/travis/.phpenv/versions/5.2.17/bin/phpunit:46
```
</details>

---

### DIFF
--- a/tests/Auth/Basic.php
+++ b/tests/Auth/Basic.php
@@ -22,11 +22,11 @@ class RequestsTest_Auth_Basic extends PHPUnit_Framework_TestCase {
 			'transport' => $transport,
 		);
 		$request = Requests::get(httpbin('/basic-auth/user/passwd'), array(), $options);
-		$this->assertEquals(200, $request->status_code);
+		$this->assertSame(200, $request->status_code);
 
 		$result = json_decode($request->body);
-		$this->assertEquals(true, $result->authenticated);
-		$this->assertEquals('user', $result->user);
+		$this->assertTrue($result->authenticated);
+		$this->assertSame('user', $result->user);
 	}
 
 	/**
@@ -43,11 +43,11 @@ class RequestsTest_Auth_Basic extends PHPUnit_Framework_TestCase {
 			'transport' => $transport,
 		);
 		$request = Requests::get(httpbin('/basic-auth/user/passwd'), array(), $options);
-		$this->assertEquals(200, $request->status_code);
+		$this->assertSame(200, $request->status_code);
 
 		$result = json_decode($request->body);
-		$this->assertEquals(true, $result->authenticated);
-		$this->assertEquals('user', $result->user);
+		$this->assertTrue($result->authenticated);
+		$this->assertSame('user', $result->user);
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


schlessera:
For the test failures `RequestsTest_Transport_cURL::testHEAD` and `RequestsTest_Transport_fsockopen::testHEAD`:

The `Requests_Response` object states that the `body` property is of type `string`, so I'd consider the fact that it returns `false` on PHP 7.0+ for the tested request a bug. Therefore, I'd suggest we detect where the `false` is happening (probably some PHP internal method producing an error, maybe `json_decode` or so?), and then catch that case an explicitly turn it into an empty string.

For the failure on `RequestsTest_Session::testURLResolution`:
Again, going by the fact that `Requests_Response` states that the `url` property is of type string, I'd consider this a bug. The `url` in this case is a `Requests_IRI` object and the loose comparison would trigger its `__toString()` method. I suggest manually calling the object's `__toString()` method (or just casting it to `(string)`) when assigning it to the `url` property.

jrfnl:
@schlessera I'll get those fixed up and send in separate PRs for those as those will constitute functional changes which need a changelog entry.

As for this PR, I've removed the three failing changes, so this should be mergable once the build has passed.

schlessera (approved):

PR #416, opened 2020-10-24
CS: use combined assignment operators

jrfnl:


ozh:
For what it's worth, I think `$i += 2;` is a no brainer but find ```$i        += 2 + $xlen;``` to be confusing. The spacing and the fact that we're so used to using this notation only to increment something with an integer.

schlessera:
I don't see an immediate issue with this notation. It might be less frequently used with an additional `+` operator, but it will still be the more performant operation here.

I personally don't like the equal sign alignment for cases like these (I think it only makes sense in large collections of assignments, not in random code), but that is what PHPCS is currently automating. If we want to change/rediscuss this, this should be done at the WPCS ruleset level.

schlessera (approved):

PR #425, opened 2020-11-05
Requests::parse_response(): ensure URL is always a string

jrfnl:
This fixes an issue on PHP 5.2/5.3 where the `Requests_Response::__toString()` method would not always automagically be invoked.

I've verified that this is the only place in the application where the `Requests_Response::$url` property is being set.

I considered making the property `private` and adding magic setter/getter methods, but that would open the class up to injection of other non-public properties, which is not the intention of this change.

Includes:
* Making the tests which cover this property use a strict typed assertion.
* Adding "failure messages" to the various assertions in this test to more easily be able to distinguish which assertion has failed.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -634,7 +634,7 @@ class Requests {
 		}
 
 		$return->raw  = $headers;
-		$return->url  = $url;
+		$return->url  = (string) $url;
 		$return->body = '';
 
 		if (!$options['filename']) {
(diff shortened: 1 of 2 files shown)


schlessera (approved):

jrfnl:
@schlessera Shall I just rebase it for a clean merge ?

schlessera:
@jrfnl Yes, you can do that. I don't know why, time and time again I think it would be a good idea to use the GitHub conflict resolution screen for small conflicts... and fail miserably at it! 🤦

jrfnl:
No worries, it happens. Rebased & force pushed now.

PR #426, opened 2020-11-05
Requests::parse_response(): ensure `body` is always a string

jrfnl:
This fixes an issue on PHP < 7.0, where the value of `Requests_Response::$body` could be `false` when there was no response body.

The behaviour of the `substr()` method has seen some (undocumented) changes across PHP versions.

In PHP < 7.0, the method would return `false` if `$start` was equal to the length of the subject string.
As of PHP 7.0, the method will return an empty string in that case and only return `false` if `$start` is _beyond_ the length of the subject string.
In PHP 8.0, this will change yet again and the function will no longer return `false` ever, but will return an empty string in all "failure" cases.

Knowing this, checking the return value of `substr()` with `empty()` is the most appropriate course of action.

I've verified that this is the only place in the application where the `Requests_Response::$body` property is being set.

Includes:
* Making the tests which cover this property use a strict typed assertion.

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -633,8 +633,9 @@ class Requests {
 			return $return;
 		}
 
-		$return->raw = $headers;
-		$return->url = $url;
+		$return->raw  = $headers;
+		$return->url  = $url;
+		$return->body = '';
 
 		if (!$options['filename']) {
 			$pos = strpos($headers, "\r\n\r\n");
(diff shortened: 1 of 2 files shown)


jrfnl:
Note: this PR and PR #425 will conflict. I'll rebase the open PR once one of the two has been merged.

schlessera (approved):

PR #428, opened 2020-11-05
Try adding support for a broader range of PHPUnit versions

TimothyBJacobs:
This adds a `RequestsTestCase` base class for all tests. This allows converting exception annotations to method calls. Previously this was only available to the transport tests.

Additionally, this handles `void` return types for the `setUp` method.


jrfnl:
@TimothyBJacobs Just checking in - do you think #436 will be a workable setup for you to work on those unit tests for the PHP 8 issues you opened ?

TimothyBJacobs:
I think so.

PR #438, opened 2020-11-22
only decompress with proper content-encoding header

### LINKED ISSUE #437: ".tgz" files somtimes decompressed in error.
Requests.php will mistakenly decompress a .tgz file (or other gzipped files) downloaded from sites that pass a "content-encoding: none" header.

In the case of .tgz files, you end up with a misnamed, uncompressed tar file.

The decompress code is wrapped by `if (isset($return->headers['content-encoding']))`, so the decompress code fires even though the encoding is "none".  Should thee be a check for a proper encoding type, or at minimum, check for "none" before decompressing.

Ran into this with bitbucket.  A sample file to demonstrate the issue: https://bitbucket.org/jerrm-bb/testdownload/downloads/outcnam-test.tgz

jerrm:
Closes #437

### DIFF
--- a/library/Requests.php
+++ b/library/Requests.php
@@ -676,7 +676,7 @@ class Requests {
 			$return->body = self::decode_chunked($return->body);
 			unset($return->headers['transfer-encoding']);
 		}
-		if (isset($return->headers['content-encoding'])) {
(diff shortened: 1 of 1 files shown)


jrfnl:
@jerrm Hiya!

Also: would you be willing and able to add some tests for this change to the PR ?

jrfnl (commented):
Some notes:
* Needs rebase & conflict solving.
* The hard-coded array should become a class constant.
* Check needs to be done whether the array is complete for these purposes.
* Can this be done without using the performance-poor `array_intersect()` ?
* Needs tests.

PR #439, opened 2020-12-04
Travis: add build against PHP 8.0

jrfnl:
PHP 8.0 has been branched off two months ago, so `nightly` is now PHP 8.1 and in the mean time PHP 8.0 was released last week.

As of today, there is a PHP 8.0 image available on Travis.

This PR adds a new build against PHP 8.0 to the matrix and, as PHP 8.0 has been released, that build is not allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ jobs:
    - php: 7.2
    - php: 7.3
    - php: 7.4
+   - php: 8.0
    - php: nightly
 
   allow_failures:


schlessera:
It looks like Travis is not being triggered anymore in this repository. I'm merging this anyway, in case Travis should resurface later.

schlessera (approved):

schlessera:
Travis did still run a test build: https://travis-ci.org/github/WordPress/Requests/builds/763695363

jrfnl:
@schlessera Looks like 7.0 failed due to the usual hickups and 8.0 as this PR did not update the condition for when to run `composer install` with `--ignore-platform-reqs`.

Other than that, happy to see everything passing.

PR #444, opened 2021-02-05
Removed CURLOPT_REFERER

twdnhfr:
Hey,

as stated in the curl docs https://everything.curl.dev/libcurl-http/requests#referrer

the Referer is a header that tells the server from which URL the user-agent was directed from.

So imho this shouldn't be the request's target url.


jrfnl:
Rebased without changes to make the PR mergable.

jrfnl:
Build failure was a hickup in CI, re-ran the build and it's passing now.

schlessera (approved):

PR #445, opened 2021-02-09
Add shebangs to proxy shell scripts

schlessera:
Without proper shebangs in the shell script, some systems assume lowest common denominator of `sh` for compatibility reasons.

This breaks with the testing logic in the current scripts:

~_dev_Requests 2021-02-09 at 9 31 46 AM

This PR adds a shebang to the scripts to use `bash` as the executable shell instead.


jrfnl (approved):
LGTM

One question: how will this behave on a non-*nix platform ?

PR #446, opened 2021-02-09
Replace PHPUnit 6+ compat code with Yoast PHPUnit polyfills package

schlessera:
This PR replaces the few pieces of PHPUnit 6+ compact code with a more strategic polyfill library (https://github.com/Yoast/PHPUnit-Polyfills).

The PR also includes changes to adapt the tests to use PHPUnit 9+ syntax as needed, which is then polyfilled down to lower PHPUnit versions.


jrfnl:
While I fully support this change, I hadn't done so myself previously as it will prohibit the tests from being run on PHP < 5.5 (PHPUnit Polyfills has a minimum support PHP version of PHP 5.5), while `Requests` officially still supports PHP 5.2, 5.3 and 5.4.

I'm totally behind dropping support for PHP < 5.6, but I think we should do that in a major, not a patch version.

With that in mind, shall we milestone this PR for 2.0.0 ?

jrfnl (changes requested):
Hi Alain, notwithstanding my other remark, I've gone through the PR anyhow to verify it.

Aside from some small remarks I've left in-line:
* **Missing**: the `RequestsTest_TestCase` class.
    All tests now extend from it, but the file itself is not included in this PR, so the tests can't be run at the moment.
    I'm presuming that that file extends the `TestCase` from PHPUnit Polyfills to actually make the polyfills available ?
* **Missing**: the PHPUnit cache file which is generated when using PHPUnit 8/9 should be added to the `.gitignore` file.

Other than that and not for this PR, but just as a reminder for later:
* The `RequestsTests_Encoding::bin2hex()` looks like it isn't used ?
* When PHP < 5.5 is officially dropped, I'd recommend namespacing the tests according to PSR4, removing the custom `autoload_tests()` function from the test bootstrap and adding an `autoload-dev` section to the `composer.json` file to handle the autoloading.
    (would also be nice for the main library, but would be a big breaking change, so that needs a different consideration)

schlessera:
> Missing: the RequestsTest_TestCase class.

Where did that one go? Seems like I forgot to add it to the commits, because I'm pretty sure this never worked without it. :)

jrfnl:
> Where did that one go? Seems like I forgot to add it to the commits, because I'm pretty sure this never worked without it. :)

It was in the branch which extended off this one with the initial GH Actions setup. (and why that branch wasn't deleted yet)

jrfnl:
I've added one significant (and one insignificant) commit which fixes a problem when running the tests against a PHPUnit phar file.

If the `composer install` had been run against a **high** PHP version, running the tests using a PHPUnit Phar file against **low** PHP versions would error out:

![image](https://user-images.githubusercontent.com/663378/122561001-923f3680-d041-11eb-97e1-de006a13bd59.png)

( `runtests` is a script which will run the tests against every PHP minor using a PHPUnit phar for the appropriate PHPUnit version)

The errors are:
```
"PHP 5.6.40"
                                                                                                                                                                                                    
Warning: Unsupported declare 'strict_types' in path/to/Requests/vendor/phpunit/phpunit/src/Framework/Assert/Functions.php on line 1                                              
                                                                                                                                                                                                    
Fatal error: Default value for parameters with a class type hint can only be NULL in path/to/Requests/vendor/phpunit/phpunit/src/Framework/Assert/Functions.php on line 89
                                                                                                                                                                                                    
"PHP 7.0.33"                                                                                                                                                                                        
                                                                                                                                                                                                    
Parse error: syntax error, unexpected '?', expecting variable (T_VARIABLE) in path/to/Requests/vendor/phpunit/phpunit/src/Framework/Assert/Functions.php on line 177
```

Note: when the `composer install` would have been run on a **low** PHP version, there would have been no issues.

All the same, by reversing commit 07b71f09fe3f0bb8e96cbddb15598cb4e11ea357 and adding a `require` for the PHPUnit Polyfills autoloader to the bootstrap, the tests can be run with a PHPUnit Phar file without problems _independently of what PHP version was used to run the `composer install`_.

jrfnl (approved):
All good now!

@schlessera I'll leave this open for the moment to give you a chance to review the additional commits I added and to test this version locally.
Happy for it to be merged though.

Issue #420, opened 2020-10-26
Switch to Travis-ci.com


jrfnl:
Travis is slowly closing down their old `.org` interface. I have seen a number of repos running into issues with Travis not reporting back on branches/pull requests anymore and sooner or later, this repo will be affected as well.

Based on my experiences with other repos, this is what needs to happen:

> How to get TravisCI working again if it's currently not reporting:
> 1. On GitHub/ repo Settings: Remove the Webhook integration.
> 2. On GitHub/ marketplace: Install the Travis GH App
> 3. On travis-ci.com: sign up for the migrations beta for the organisation
> 4. On travis-ci.com: migrate all relevant repos
> 5. For PRs without status: force push problem PRs (no change needed in the content of the PR)
> 6. Once a first build has finished under the new system, change the protected branches to require "Travis CI - Pull Request" and un-require "continuous integration"

Source: https://twitter.com/jrf_nl/status/1255854838597984262

As I don't have admin rights for this repo, I can't action this myself. @rmccue Would you please be so kind to action this ?

jrfnl:
Let's put a pin in this for now. While travis.org is dreadfully slow at the moment, the .com interface is forcing paid plans since this week and the option for "free builds for Open Source" requires a request to be added to that plan, except they "forgotten" to build the interface to request this... *sigh*

Probably better to look at switching to GH actions in the near future.

ozh:
My 2 cents : while switching to GH actions is certainly the sound thing to do, migrating to travis-ci.com litterally takes 15 seconds and works quite well (my experience these past 3 days at least). There's no reason to postpone, really.

jrfnl:
@ozh Not in my experience. Did so on various projects and with the limited open source/free tokens, my builds would stop running by day three of the month, so Travis is no longer a viable options.

ozh:
Well, not sure I still get it... Travis.org is 100% not working, Travis.com maybe working, so it's still better than nothing for a 15 seconds investment :)

schlessera:
We're making good progress moving everything to GHA, so I'm closing this issue.

PR #447, opened 2021-02-09
Replace Travis CI with GitHub Actions workflows

schlessera:


.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+
+  lint: #-----------------------------------------------------------------------
+    name: Lint PHP files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+
+      - name: Set up PHP envirnoment
```
jrfnl, line 18:
```suggestion
      - name: Set up PHP environment
```
(here and everywhere else)

jrfnl:
Had a quick look through what is here so far (this is not a full review).

* You may want to have a look at the [`ramsey/composer-install`](https://github.com/ramsey/composer-install) action - this will save on a lot of duplication with the composer install/composer caching.
* IMO the `lint` action should be run - at least - on the lowest + highest supported PHP version.
    For a library like this which is used as a dependency for other projects, I'd even go so far as to run it on all supported PHP versions just to be sure.
* Pro-tip: for the `lint` action, exclude the `.git` directory.
    This is not excluded by default in Parallel Lint and for "old" projects, not excluding it can really slow the action down _a lot_ as the script will search the whole `.git` directory for PHP files.
    Oh and for this repo, you may also want to exclude the `docs` directory, though that shouldn't make much difference.
* Suggestion: add a `lint` script to the `composer.json` file to make it easy for people to run the linting locally with the correct parameters:
    ```json
    "lint": [
      "@php ./vendor/php-parallel-lint/php-parallel-lint/parallel-lint . -e php --exclude vendor --exclude .git
    ],
    ```
    Note:
    * the `@php` is to make sure the same PHP version is used as used by Composer (instead of a default PHP version on a system);
    * the full path name is for improved compatibility cross-platform (i.e. make sure it also works on Windows).
* I hadn't come across the `cs2pr` usage before, but if it does what I expect it to do, I **_like_** it 👍 
    Will have to nick that trick and apply it to various workflows in other projects 😉 
* I generally prefer to run the QA checks against all pushes, not just `master`, though for the tests - depending on how complex the matrix is - I sometimes use a simpler, only high/low job for pushes.
    Providing people check the action outcome before opening a PR, it should prevent PRs being opened which don't pass on the basics, which then have to get a second commit/force push to fix the PR as soon as it's opened.
* You can add PHP `8.1` to the matrix for the `unit` job to test against the current PHP `nightly`.
    Use it in combination with `continue-on-error` to prevent it from failing/stopping the workflow run.
* Once the script is finished, you may want to remove the `fail-fast: false` to be kind to the GH resources.

jrfnl:
Just making a note here that this PR would close #420 + #439

jrfnl:
Oh and another reminder: let's not forget to update the badges in the `Readme` once this PR is ready.

jrfnl:
I'm moving this back to the 2.0.0 milestone as we do still seem to be able to view the Travis runs, which allows us to release 1.8.0 knowing that CI is passing, so we can leave this until after the 1.8.0 release.

schlessera:
Closing this for now, as it is being superseded by #467

Issue #315, opened 2018-02-17
Update httpbin endpoint on tests


soulseekah:
The default `REQUESTS_TEST_HOST` endpoint (requests-php-tests.herokuapp.com) points to a very outdated version of httpbin forcing a couple of tests to be run directly against https://httpbin.org instead.

The tests are:

7b099c59b152f746c0ea1324f138eab122f76c0c - /status/307 redirect issue fixed in later versions (see #7)
3ebc94a1cd5526e19b072ad49c023e0b3fe984f0 - /post does not show form data and file data at the same time in older versions

Any chance that app can be updated so tests can use the `httpbin()` function in the newer cases?

rmccue:
Just noting that the Heroku app is *not* the same as httpbin.org, but rather a custom one which runs https://github.com/RequestsPHP/test-server - this exists because httpbin.org doesn't support all the tests we need.

If httpbin.org does now support those, I'd be very up for removing that custom server, as it's not great.

carlalexander:
Just wanted to leave a note that the tests I'm writing right now for #454 only seem to work against the test server when it's run locally. Heroku seems to strip the header I'm testing against. I've also tried httpbin.org. So I don't think we can remove the test server at this time.

ozh:
After switching to Github Actions, it should be a breeze to simply `docker run -p 80:80 kennethreitz/httpbin` and run tests locally

rmccue:
FYI, "This app is using the Cedar-14 stack, which reached end-of-life on May 1st, 2019" - not sure if this has been migrated yet. It'll likely go offline whenever Heroku makes their change to remove free dyno support too.

jrfnl:
@rmccue Not sure where that message is coming from, but we replaced Heroku with Render nearly two years ago as their free support stopped. See PR #784/#815

rmccue:
Heroku's been emailing me - I had a feeling it'd been moved, but wanted to double-check just in case it'd break something :)

jrfnl:
@rmccue Ah, I see. Thanks for the heads-up.

PR #454, opened 2021-03-12
Manage "Expect" header with cURL transport

### LINKED ISSUE #453: Performance issue with "Expect" header and cURL
By default, cURL adds a `Expect: 100-Continue` header to certain requests. This can add as much as a second delay to requests done using cURL. This is [discussed on the cURL mailing list](https://curl.se/mail/lib-2017-07/0013.html).

The general practice is to set an empty `Expect` header when the request payload is small. This is something other cURL clients like [Guzzle](https://github.com/guzzle/guzzle/blob/d5a2b0338e1cc963532ffdcc3bba4cbba469a215/src/RequestOptions.php#L96-L112) and [Stripe](https://github.com/stripe/stripe-php/blob/3febf22da8ed6a3b45842eac97d0fcece4d91e13/lib/HttpClient/CurlClient.php#L244-L256) do.

carlalexander:
Fixes #453

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,7 @@
+codecov:
+    notify:
+        after_n_builds: 2
+
 coverage:
   round: nearest
   # Status will be green when coverage is between 85 and 100%.
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+.github/ export-ignore
 bin/ export-ignore
 docs/ export-ignore
 examples/ export-ignore
--- a/.github/workflows/cs.yml
+++ b/.github/workflows/cs.yml
@@ -1,0 +1,37 @@
+name: CS
+
+on:
+  # Run on all pushes and on all pull requests.
+  push:
+  pull_request:
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  phpcs: #----------------------------------------------------------------------
+    name: 'PHPCS'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          coverage: none
+          tools: cs2pr
+
+      # Install dependencies and handle caching in one go.
+      # @link https://github.com/marketplace/actions/install-composer-dependencies
+      - name: Install Composer dependencies
+        uses: "ramsey/composer-install@v1"
+
+      # Check the code-style consistency of the PHP files.
+      - name: Check PHP code style
+        continue-on-error: true
+        run: composer checkcs -- --report-full --report-checkstyle=./phpcs-report.xml
+
+      - name: Show PHPCS results in PR
+        run: cs2pr ./phpcs-report.xml
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,45 @@
+name: Lint
+
+on:
+  # Run on all pushes and on all pull requests.
+  push:
+  pull_request:
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  lint: #----------------------------------------------------------------------
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        # Lint against the high/low versions of each PHP major and select additional versions.
+        # Note: linting against PHP 5.2 , 5.3 and 5.4 is still done via Travis.
+        php: ['5.6', '7.0', '7.4', '8.0']
+        experimental: [false]
+
+        include:
+          - php: '8.1'
+            experimental: true
+
+    name: "Lint: PHP ${{ matrix.php }}"
+    continue-on-error: ${{ matrix.experimental }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          tools: cs2pr
+
+      # Install dependencies and handle caching in one go.
+      # @link https://github.com/marketplace/actions/install-composer-dependencies
+      - name: Install Composer dependencies
+        uses: "ramsey/composer-install@v1"
+
+      - name: Lint against parse errors
+        run: composer lint -- --checkstyle | cs2pr
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -1,0 +1,96 @@
+name: Quicktest
+
+on:
+  push:
+    branches-ignore:
+      - master
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  #### QUICK TEST STAGE ####
+  # Runs the tests against select PHP versions for pushes to arbitrary branches.
+  quicktest:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php: ['5.5', '7.2', 'latest']
+
+    name: "Quick Test: PHP ${{ matrix.php }}"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+
+      # Install dependencies and handle caching in one go.
+      # @link https://github.com/marketplace/actions/install-composer-dependencies
+      - name: Install Composer dependencies - normal
+        if: ${{ startsWith( matrix.php, '8' ) == false && matrix.php != 'latest' }}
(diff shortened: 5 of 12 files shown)


carlalexander:
I'm not sure how you prefer to handle the `strlen` calculation since `data` can be an array or a string. I did it inside the `strlen` happy to change it to something else.

jrfnl:
@carlalexander Thanks for finding this and for the PR!

Do you think this could be covered by unit tests in some way ?

> I'm not sure how you prefer to handle the strlen calculation since data can be an array or a string. I did it inside the strlen happy to change it to something else.

For code readability and comprehension, I'd much prefer it if it would be pulled apart and calculated before the `strlen()`, but that's just me.

Also, for the multi-line comment, you may want to use multi-comment style `/*... */` ;-)

carlalexander:
Normally, I do multi-line comments that way too. I'm just trying to keep the same coding style as the one below. Happy to switch it to the `/*`.

I can also create a function to get the get the `strlen`.

For testing, I'd like to add tests. I just didn't really see anything for testing cURL code. The cURL transport has 4 tests to check for some thrown exceptions. That's it.

jrfnl:
> For testing, I'd like to add tests. I just didn't really see anything for testing cURL code.

I see where your confusion comes from. Most of the tests for the cURL code are in the `RequestsTest_Transport_Base` test class which the `RequestsTest_Transport_cURL` test class extends.
The `RequestsTest_Transport_cURL` class only contains the cURL specific tests, so I imagine a test for this cURL specific issue would be best placed in that file.

If you run the tests with `phpunit --testdox --filter RequestsTest_Transport_cURL`, you can see that the cURL class is tested with a lot more tests than just those in the one file.

jrfnl:
> Normally, I do multi-line comments that way too. I'm just trying to keep the same coding style as the one below. Happy to switch it to the /*.

I hear you. We've cleaned up the code style recently, cleaning up the comments is on the to do list still, ;-)

carlalexander:
Alright, I'll take a look at adding tests over the weekend. 

I've done another pass and fixed up the comment and moved the `strlen` code to its own method. I added the method at the end because I can't seem to figure out how methods are ordered. Just let me know where you'd like it.

carlalexander:
Alright, so I've tried to write some tests, but the Heroku application that we run tests against doesn't seem to return the `Expect` header back. I've tried with both an empty header and one with `100-Continue`. 

Is the code for that Heroku application anywhere? Also it could be that Heroku strips the header.

jrfnl:
@carlalexander I'm not sure at all, but I suspect it might be this repo: https://github.com/rmccue/httpbin

@rmccue Ryan, could you please confirm ? (or point to the real source)

carlalexander:
Yeah, I'm trying against `https://httpbin.org/headers` and it's not returning the header regardless that its set or not.

schlessera:
I think the source of the server is this one: https://github.com/RequestsPHP/test-server

rmccue:
> I think the source of the server is this one: https://github.com/RequestsPHP/test-server

That's the one, yep :)

jrfnl:
@carlalexander Just checking in - will you have time to update the PR for the tests ? And if so, what timeline are we looking at ?

carlalexander:
@jrfnl Hoping to sit down and hammer at it some more this weekend. 😄  I haven't looked at it this week.

carlalexander:
Ok, so I have some tests running, but they only work with @rmccue's server. I'm just running it locally atm. Are there tests that use it? It seems everything uses httpbin.org and it doesn't return the header. 

Another good news is you can clearly see the 1 sec timeout with the tests. So that's great.

image

jrfnl:
@carlalexander I hope I understood your question correctly, but AFAICS, all tests which use a function call to `httpbin()` to set the address use the custom test server.

Also see #315 for some additional context.

carlalexander:
Ok, this helped @jrfnl! I tried with `requests-php-tests.herokuapp.com` and it also strips the header. But Travis seems to run the tests against the test server running with the built-in PHP server. So I pushed the tests up and they seem to work on Travis!

jrfnl:
@carlalexander Awesome! I'll have a more detailed look soon. What about annotating the tests with the performance fix with [`@small`](https://phpunit.readthedocs.io/en/9.5/annotations.html#small) to ensure they fail if the extra second would be added ?

jrfnl:
@carlalexander Was just looking at Travis and I notice there does still seem to be an issue:
```
1) RequestsTest_Transport_cURL::testGETWithNestedData

Array to string conversion

/home/travis/build/WordPress/Requests/library/Requests/Transport/cURL.php:573
/home/travis/build/WordPress/Requests/library/Requests/Transport/cURL.php:332
/home/travis/build/WordPress/Requests/library/Requests/Transport/cURL.php:135
/home/travis/build/WordPress/Requests/library/Requests.php:381
/home/travis/build/WordPress/Requests/tests/Transport/Base.php:111

2) RequestsTest_Transport_cURL::testPOSTWithNestedData

Array to string conversion

/home/travis/build/WordPress/Requests/library/Requests/Transport/cURL.php:573
/home/travis/build/WordPress/Requests/library/Requests/Transport/cURL.php:332
/home/travis/build/WordPress/Requests/library/Requests/Transport/cURL.php:135
/home/travis/build/WordPress/Requests/library/Requests.php:381
/home/travis/build/WordPress/Requests/library/Requests.php:270
/home/travis/build/WordPress/Requests/tests/Transport/Base.php:222
```

carlalexander:
@jrfnl That's my bad. I was using the new function signature for `implode`. Fixed it. Also added `@small` tags.

carlalexander:
Ok, well the old signature makes `phpcs` fail now lol

jrfnl:
> using the new function signature for implode.

There is no "new" function signature. There is a cross-version compatible signature and a second signature which was previously supported, but isn't anymore.

The function should be called like so: `implode ( string $separator , array $array )`

> Ok, well the old signature makes phpcs fail now lol

And rightfully so as using the reverse order will throw a deprecation notice on PHP 7.4 (and fail the tests because of it) and be a fatal error on PHP 8.0.

carlalexander:
Yeah, I'm reverting the change. I understand what's going on now. I'm also going to add some more tests.

jrfnl:
@carlalexander Appreciated! You're a ⭐

carlalexander:
Everything is green except for PHP 8.0. It doesn't seem to be an issue with the tests or code. It seems to be PHPUnit related.

jrfnl:
> Everything is green except for PHP 8.0. It doesn't seem to be an issue with the tests or code. It seems to be PHPUnit related.

Correct - see #439

carlalexander:
Perfect, should be good to merge once you've done your review. Let me know if you need anything else 😄

library/Requests/Transport/cURL.php
```diff
@@ -546,4 +561,23 @@ public static function test($capabilities = array()) {
...
+		if (is_array($data)) {
+			$iterator = new RecursiveIteratorIterator(new RecursiveArrayIterator($data));
+			$data     = '';
+
+			foreach ($iterator as $datum) {
+				$data .= $datum;
+			}
+		}
+
+		return strlen($data) >= 1048576 ? '100-Continue' : '';
```
schlessera, line 581: This is not very efficient given we're only interested in the byte size of the array. The concatenation of the string can be expensive in terms of memory operations, and then you still haven't counted the bytes yet.

How about this instead:
```suggestion
		if (is_array($data)) {
			$iterator = new RecursiveIteratorIterator(new RecursiveArrayIterator($data));
			$bytesize = 0;

			foreach ($iterator as $datum) {
				$bytesize += strlen((string)$datum);
			}
		} else {
			$bytesize = strlen((string)$data);
		}

		return $bytesize >= 1048576 ? '100-Continue' : '';
```

schlessera (changes requested):

carlalexander:
Pushed @schlessera's recommended changes 😄

library/Requests/Transport/cURL.php
```diff
@@ -546,4 +561,25 @@ public static function test($capabilities = array()) {
...
+	 * @param string|array $data Data to send either as the POST body, or as parameters in the URL for a GET/HEAD.
+	 * @return string The "Expect" header.
+	 */
+	protected function get_expect_header($data) {
+		if (is_array($data)) {
+			$iterator = new RecursiveIteratorIterator(new RecursiveArrayIterator($data));
+			$bytesize = 0;
+
+			foreach ($iterator as $datum) {
+				$bytesize += strlen((string) $datum);
```
TimothyBJacobs, line 577: I think it might be worth bailing out here immediately once the `bytesize` limit is reached.

carlalexander:
Really great suggestion @TimothyBJacobs! I've reworked the function to early return if we hit the 1MB limit.

schlessera:
Dammit! I broke the PR when I wanted to do a rebase. When will I learn not to touch PRs that were done directly in `master`?

@carlalexander Are you able to push the change again rebased on latest `master`? If not, let me know, and I'll recreate your PR.

schlessera:
I recreated the PR here: #469

PR #457, opened 2021-03-20
FilteredIterator: fix PHP 5.2/5.3 compatibility and actually run the tests

jrfnl:
### FilteredIterator: fix PHP 5.2 compatibility and stabilize the code

The `$callback` passed to the constructor of the `FilteredIterator` class may not be callable and if so, would cause a fatal error.

This puts some defensive coding in place to guard against that and adds a unit test for it as well.

This also fixes the tests failing on PHP 5.2, as when the object is unserialized on PHP 5.2, the `$callback` property will no longer be set. This looks to have been a bug in PHP 5.2 which was fixed in PHP 5.3.

### Tests/FilteredIterator: fix a parse error on PHP 5.3

Even though the code is wrapped in a version check for running, the code still needs to parse correctly on low PHP versions.

This was not the case now, which could break the running of the tests of PHP 5.3.

Fixed now.

### Tests: actually run the test for the FilteredIterator

As all tests are explicitly named in the test configuration, the  test for the `FilteredIterator` as added in #422 were not being run as they hadn't been added to the config.

This fixes that.

### DIFF
--- a/library/Requests/Utility/FilteredIterator.php
+++ b/library/Requests/Utility/FilteredIterator.php
@@ -39,7 +39,11 @@ class Requests_Utility_FilteredIterator extends ArrayIterator {
 	 */
 	public function current() {
 		$value = parent::current();
-		$value = call_user_func($this->callback, $value);
+
+		if (is_callable($this->callback)) {
+			$value = call_user_func($this->callback, $value);
+		}
+
 		return $value;
 	}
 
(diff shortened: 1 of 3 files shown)


jrfnl:
I've added a second commit to enable actually running the test....

jrfnl:
The build against [PHP 5.2 is failing for this PR](https://travis-ci.org/github/WordPress/Requests/jobs/763741271) and I'm wondering whether to spend time on fixing the test for this or to just skip this particular test on PHP 5.2.

@schlessera Opinion ?

jrfnl:
Okay, fixed it as the fix which is needed for 5.2, actually makes sense for all PHP versions anyhow.

So I've added (yet another) commit and updated the PR description to cover all three changes.

jrfnl:
This is now ready for merge - passing build: https://travis-ci.org/github/WordPress/Requests/builds/763756355

PHP 8 is failing as discussed in #439. Doesn't really matter as `nightly` is currently `8.0.3-dev` and passing.

schlessera (approved):

PR #461, opened 2021-03-21
Travis: fix PHP 8.0 build

jrfnl:
Follow up on #439

Even though we'll soon be moving to GH Actions, I think it is important to fix the failing build on PHP 8.0 ahead of this and ahead of the 1.8.0 release, as otherwise any badges in the README, on Packagist and on the website, will show the CI build as failing until we release version 2.0.0 and that wouldn't be right.


jrfnl:
Build is 🟢 : https://travis-ci.org/github/WordPress/Requests/builds/763768465

schlessera (approved):

PR #458, opened 2021-03-20
Move the PHPUnit config file to the project root

jrfnl:
Just simplifies things and makes the location more standard/predicatable.

PR #467, opened 2021-03-22
CI: switch to GitHub Actions for PHP 5.5 - current

jrfnl:
Depends on: PRs #458 (and by extension PR #457) and #464.
Those PRs will need to be merged before this PR can become mergable.

Supersedes: PRs #447 and #461

Closes #447
Closes #461

## TL;DR

This PR moves the CI checks for PHP 5.5 - current to GH Actions, while leaving the CI checks against PHP 5.2 - 5.4 to be run via Travis. This allows for this move to be done _now_ while PHP 5.2 - 5.4 are still supported and will still safeguard (via Travis) that the code is still compatible with PHP 5.2 - 5.4.

This PR incorporates the work previously done by @schlessera in #447 and builds on that.

**Background of why the PHP 5.2-5.4 builds won't be able to move to GH Actions:**
The code of the test-server used during the running of the unit tests requires PHP 5.5 minimum due to the use of `yield`.

On Travis, it is possible to use PHP 5.6 to start and run the test server, while running the tests themselves on a lower PHP version within the same workflow.

On GH Actions, while it is possible to switch PHP versions during a workflow, this switch will affect all code, so the test server can not run on a different PHP version than the tests, effectively blocking PHP 5.2-5.4.

### CI: switch to GitHub Actions - step 1: sniff/CS workflow

This commit:
* Adds a GH Actions workflow for the CI checks which were previously run on Travis in the `sniff` stage.
* Removes that part of the `.travis.yml` configuration.
* Adds a "Build Status" badge in the Readme to use the results from this particular GH Actions runs.

### CI: switch to GitHub Actions - step 2: lint workflow

This commit:
* Add a new dependency on the PHP Parallel Lint package which will be used for linting on PHP >= 5.4.
    The PHP Parallel Lint package, in combination with the PHP Console Highlighter provides the following advantages in comparison with "plain" PHP linting:
    - Higher speed due to the parallel processes.
    - Improved usability by providing color coded syntax highlighting of found errors on the command-line.
    - Integration with the `cs2pr` tool, allowing for the results of the lint command to be shown in-line in PRs.
* Adds a Composer `lint` script for easy access to the tool for devs, while making sure the correct command line parameters will be used.
* Adds a GH Actions workflow for linting the code on the high/low minors of each PHP major version.
    Note: this workflow will only run on PHP >= 5.4. Linting against PHP 5.2 and 5.3 will remain in Travis for now.
* Removes the part of the `.travis.yml` configuration which has now been replaced.
* Adds a "Build Status" badge in the Readme to use the results from this particular GH Actions runs.

### CI: switch to GitHub Actions - step 3: test and coverage stage

This commit:
* Adds a GH Actions workflow for running the unit tests against PHP 5.5 - current.
    Note:
    - This workflow will run for all pull requests and for merges to master, but not for other push events.
    - Code coverage will be checked for the highest and lowest PHP version so the results can be combined to cover any polyfills and work-arounds.
        Includes adjusting the codecov configuration to expect 2 reports for each build.
    - The PHP 8.1 (nightly) job is "allowed to fail" (`experimental` = true). If it does fail, the build status will unfortunately show as "failed", even though all non-experimental jobs have succeeded.
         This is a known issue in GHA: actions/toolkit#399
    - For PHP 5.2-5.4, the tests will still be run via Travis as the test server setup is incompatible with PHP < 5.5 and using multiple PHP versions in a workflow run does work on Travis, but doesn't work the same on GH Actions.
* Strips down the `.travis.yml` configuration to the bare minimum to run the tests against PHP 5.2 - 5.4.
* Adds a "Build Status" badge in the Readme to use the results from this particular GH Actions runs.

Note: for the time being, the tests will not be run against PHP 8.1/nightly as PHPUnit 7 is incompatible, meaning that that build would _always_ fail.
Once the version constraints for PHPUnit have been widened (see PR #446), the build against PHP nightly should be enabled.

### CI: switch to GitHub Actions - step 4: introduce a quick test workflow

Previous, the Travis all-in-one workflow would run on every push and every PR, making for a lot of redundancy in the builds.

GH Actions allows to run workflows far more selectively.

The "Test" workflow has been configured to run against all supported PHP versions for pushes (merges) to `master`, as well as for pull requests.

This commit introduces a "Quick test" workflow, which is basically the same as the full "Test" workflow (without code coverage checking).
This workflow will only run the tests against a few select PHP versions on `push` events, when the push is not to the `master` branch.

This allows for people working on feature branches to still get a quick impression of whether their branch is ready to be pulled, while conserving resources by not running the full build until the branch has been pulled.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -1,0 +1,96 @@
+name: Quicktest
+
+on:
+  push:
+    branches-ignore:
+      - master
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  #### QUICK TEST STAGE ####
+  # Runs the tests against select PHP versions for pushes to arbitrary branches.
+  quicktest:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php: ['5.5', '7.2', 'latest']
+
+    name: "Quick Test: PHP ${{ matrix.php }}"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+
+      # Install dependencies and handle caching in one go.
+      # @link https://github.com/marketplace/actions/install-composer-dependencies
+      - name: Install Composer dependencies - normal
+        if: ${{ startsWith( matrix.php, '8' ) == false && matrix.php != 'latest' }}
+        uses: "ramsey/composer-install@v1"
+
+      # For PHP 8.0 and "nightly", we need to install with ignore platform reqs.
+      - name: Install Composer dependencies - with ignore platform
+        if: ${{ startsWith( matrix.php, '8' ) || matrix.php == 'latest' }}
+        uses: "ramsey/composer-install@v1"
+        with:
+          composer-options: --ignore-platform-reqs
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Setup proxy server
+        run: pip3 install mitmproxy
+
+      - name: Start test server
+        run: |
+          PORT=8080 vendor/bin/start.sh
+          echo "REQUESTS_TEST_HOST_HTTP=localhost:8080" >> $GITHUB_ENV
+
+      - name: Start proxy server
+        run: |
+          PORT=9002 tests/utils/proxy/start.sh
+          PORT=9003 AUTH="test:pass" tests/utils/proxy/start.sh
+          echo "REQUESTS_HTTP_PROXY=localhost:9002" >> $GITHUB_ENV
+          echo "REQUESTS_HTTP_PROXY_AUTH=localhost:9003" >> $GITHUB_ENV
+          echo "REQUESTS_HTTP_PROXY_AUTH_USER=test" >> $GITHUB_ENV
+          echo "REQUESTS_HTTP_PROXY_AUTH_PASS=pass" >> $GITHUB_ENV
+
+      - name: Ensure the HTTPS test instance on Heroku is spun up
+        run: curl -s -I http://requests-php-tests.herokuapp.com/ > /dev/null
+
+      - name: Check mitmproxy version
+        run: mitmdump --version
+
+      - name: Ping localhost domain
+        run: ping -c1 localhost
+
+      - name: Access localhost on port 8080
+        run: curl -i http://localhost:8080
+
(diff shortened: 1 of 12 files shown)


jrfnl:
Note: these workflows will automatically be disregarded for the `gh-pages` branch and pull requests against that branch as the workflows _don't exist_ in that branch.

jrfnl:
FYI: The linting failure against PHP 8.1 as can currently be seen, is unrelated to this repo. I've opened an issue about this with the `setup-php` action: https://github.com/shivammathur/setup-php/issues/437

jrfnl:
FYI: The issue in the `setup-php` action which I identified has been fixed in the mean time 🎉

.github/workflows/test.yml
```diff
@@ -0,0 +1,145 @@
...
+      # - coverage: Whether to run the tests with code coverage.
+      # - experimental: Whether the build is "allowed to fail".
+      matrix:
+        php: ['5.6', '7.0', '7.1', '7.2', '7.3', '7.4']
+        coverage: [false]
+        experimental: [false]
+
+        include:
+          # Run code coverage on high/low PHP.
+          # Note: use PHP 7.4 as "high" as PHPUnit 7 in combination with PHP 8 will fail to run coverage (NEEDS CHECKING)
```
schlessera, line 27: The setup seems to use 8.0 now. Does this mean this was checked and the note is out-of-date?

.github/workflows/quicktest.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+
+      # Install dependencies and handle caching in one go.
+      # @link https://github.com/marketplace/actions/install-composer-dependencies
+      - name: Install Composer dependencies - normal
+        if: ${{ matrix.php < 8.0 }}
```
schlessera, line 35: I assume this condition evaluates to `false` on `'latest'`?

.github/workflows/test.yml
jrfnl: Correct, removing it now.

.github/workflows/quicktest.yml
jrfnl: Actually, let me stabilize those conditions a little more by using `startsWith()`

jrfnl: (but yes, those did work otherwise the QuickTest workflow would already have been failing)

jrfnl:
Fixed up in the existing commits and force pushed (only changes those mentioned above).

schlessera (approved):

schlessera (approved):

PR #473, opened 2021-04-25
Use direct download of proxy to fix Travis

schlessera:
This PR skips using `pip` (which is broken on old Travis images) to install the proxy and instead downloads it directly from https://mitmproxy.org/downloads/.

It also adapts the syntax of the `ps` command in the `start.sh` script for the proxy:

https://travis-ci.org/github/WordPress/Requests/jobs/767860940#L425

Finally, it adds a Travis-specific `start-travis.sh` file to launch the old binary needed on these images with the old arguments that don't work with recent versions of the proxy anymore.

Once we got rid of the Travis CI integration, the `tests/utils/proxy/start-travis.sh` file can be removed again.


jrfnl:
Travis still appears to be broken, so it looks like this (alone) does not solve it.

.travis.yml
```diff
@@ -5,11 +5,8 @@ jobs:
   fast_finish: true
   include:
    - php: 5.2
-     dist: precise
```
jrfnl, line 8: These changes will definitely not work as Xenial doesn't support those PHP versions.

schlessera:
Travis passes again with the above changes:

Build #1143 - WordPress_Requests - Travis CI - Google Chrome 2021-04-26 at 7 50 08 AM

https://travis-ci.org/github/WordPress/Requests/builds/768379650

jrfnl:
Hero! 🎉 

Only one problem left: while the build was passing, the tests weren't actually being run. This was due to the `$PHPUNIT_BIN` not being defined since the GH Actions merge (mea culpa). I've added a commit which should hopefully fix this. 🤞

jrfnl:
Tests are running and we have one failing test to fix.

jrfnl:
Hmm.. actually two - one on PHP 5.2-5.4 and one for PHP 8.0 by the looks of it.

jrfnl:
We have lift-off 🚀 

My last commit fixed the PHP 5.2-5.4 failure and the PHP 8.0 test failure has magically disappeared.

Shall we squash this in one atomic commit before merging ?

jrfnl (approved):

schlessera:
Yay! We'll need to enable squash merges first, it seems...

PR #483, opened 2021-04-27
Add a release checklist

jrfnl:
... to document the release process and as a reminder for the future.

### DIFF
--- a/.github/release-checklist.md
+++ b/.github/release-checklist.md
@@ -5,6 +5,7 @@ Template to use for release PRs from `develop` to `master`
 
 PR for tracking changes for the x.x.x release. Target release date: **DOW MONTH DAY YEAR**.
 - [ ] Check if any dependencies need updating.
+- [ ] Update the version constant in `library/Requests.php`.
 - [ ] Add changelog for the release - PR #xxx
 - [ ] Merge this PR.
 - [ ] Make sure all CI builds are green.
--- a/.github/release-checklist.md
+++ b/.github/release-checklist.md
@@ -1,0 +1,22 @@
+Template to use for release PRs from `develop` to `master`
+===========================================================
+
+:warning: **DO NOT MERGE (YET)** :warning:
+
+PR for tracking changes for the x.x.x release. Target release date: **DOW MONTH DAY YEAR**.
+- [ ] Check if any dependencies need updating.
+- [ ] Add changelog for the release - PR #xxx
+- [ ] Merge this PR.
+- [ ] Make sure all CI builds are green.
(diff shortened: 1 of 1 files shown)


ozh:
Does this file then get automatically displayed somewhere as a prefilled template ?

jrfnl:
> Does this file then get automatically displayed somewhere as a prefilled template ?

No, it does not. Unless it has changed without me noticing, AFAIK you can have multiple issue templates, but only one (1) PR template.
With that in mind, it's not useful to set it up as a PR template as it would just confuse people, but having it available in the `.github` directory still makes it easy to access it and copy/paste the content to a release issue for maintainers.

And independently of the template part of it, it is also a way to just document the steps for doing a release.

schlessera:
For WP-CLI, I've added the release check list as an issue template which can be generated (see the last two entries): https://github.com/wp-cli/wp-cli/issues/new/choose

jrfnl:
@schlessera Will have a look and update over the weekend.

ozh:
You guys have seen this : https://probot.github.io/apps/release-drafter/ ?

jrfnl:
@ozh My experience with any type of changelog generators is absolutely lousy. They create technical changelogs based on PR titles, which in most cases are completely useless for end-users.

Case in point from this week: https://github.com/gitextensions/gitextensions/blob/release/3.5/GitUI/Resources/ChangeLog.md

That release contained at least two breaking changes ([gitextensions#9112](https://github.com/gitextensions/gitextensions/issues/9112) and [gitextensions#9113](https://github.com/gitextensions/gitextensions/issues/9113)). Aside from the fact that those shouldn't have been allowed to go into a minor, they also weren't mentioned in the changelog and finding the PR which caused them meant having to go through 100+ PRs to see if I could figure it out. I found the cause of one, but not the other.

Yes, it's a pain in the neck having to write up a changelog, but the pain in the neck for end-users being confronted by an absolutely useless PR list is much larger.

jrfnl:
> For WP-CLI, I've added the release check list as an issue template which can be generated (see the last two entries): 

@schlessera I've had a look at this, but that would make it an issue template, while it is (was) intended as a PR template for the `develop` to `master` merge PR.

ozh:
FWIW, here's how I do it for YOURLS (infrequent releases so I was always forgetting what's to be done) : I have a draft release that I use as a template and copy/paste everytime then fill, with inline HTML comment helper. Impossible to miss since it's the first thing I see on the relase page when I want to tag a new one :)

Release page:

image

Source : 

image

schlessera (approved):

PR #484, opened 2021-04-29
Add misc. files to gitattributes

mbabker:
There are a handful of repository specific config and info files still downloaded with a fresh `composer install` of this package, these are now ignored

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,7 +9,5 @@ tests/ export-ignore
 .gitattributes export-ignore
 .phpcs.xml.dist export-ignore
 .travis.yml export-ignore
-CHANGELOG.md export-ignore
 package.xml.tpl export-ignore
 phpunit.xml.dist export-ignore
-README.md export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,9 +3,13 @@ bin/ export-ignore
 docs/ export-ignore
 examples/ export-ignore
 tests/ export-ignore
+.codecov.yml export-ignore
 .coveralls.yml export-ignore
 .gitignore export-ignore
 .gitattributes export-ignore
 .phpcs.xml.dist export-ignore
 .travis.yml export-ignore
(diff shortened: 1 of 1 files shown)


jrfnl:
Hi @mbabker Good catch! Yes, we've been moving some files around and clearly hadn't kept the `.gitattributes` file completely in sync.

I agree on the CodeCov and PHPUnit config file entries. I think the `.coveralls.yml` entry can actually be removed from the file as the file no longer exists in the repo.

As for the README and the CHANGELOG: I would normally keep let those be available in the distribution package, if for no other reason than that the README contains the installation/first steps for using information and the CHANGELOG shows the version which was downloaded. Happy to consider different point of view about this though.

mbabker:
Personally, I rarely crawl through my `vendor` directories looking for the changelog or readme files.  Generally, my workflow is to run `composer outdated` and check the changelogs on affected packages before giving the update command (if you're reviewing the changelog after applying the update, you might already be in a less-than-ideal state, especially around major upgrades).  I actually only noticed this one because I added Requests to an application that isn't otherwise Composer based and didn't feel like messing with ZIP packages or manually defined autoloaders or any of the other hassles that Composer already solves to bring it in, and when checking in the `vendor` folder that's when I noticed the files.

jrfnl:
I hear you, but the `.gitattributes` file doesn't only apply to Composer updates, it also applies to the `zip` and `tar` files and while you've now made the choice to use Composer, we are aware that not all users of this library update the package that way.

mbabker:
That's fair.  I'd still say you should be reviewing those before download, but I'm known to be opinionated :wink:

jrfnl:
Appreciated! Thank you.

We'll be starting to merge things for 2.0.0 in a few weeks (just in case a 1.8.x would still be needed). I've earmarked this PR for that first merge round.

schlessera (approved):

PR #485, opened 2021-05-11
Requests::Version: update version number

jrfnl:
Oops.. looks like we forgot something for the release. Only just noticed it when I checked the file diff for the WP Core patch.

As the version constant is generally used to verify the version for using certain features of a package, does this warrant a `1.8.1` release ?

And if so, should I update the version number to `1.8.1` in this PR ?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,9 +4,8 @@ docs/ export-ignore
 examples/ export-ignore
 tests/ export-ignore
 .codecov.yml export-ignore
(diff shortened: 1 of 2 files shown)


schlessera (approved):

jrfnl:
For anyone coming upon this after the fact: This was discussed off GH and the decision was made that - as we don't know whether there are integrations which rely on the version number, we should play it safe -, so I updated the PR and we have since released version 1.8.1.

PR #491, opened 2021-06-18
Add .editorconfig file

schlessera:
This PR adds an `.editorconfig` file to let the IDE know how to handle newlines, tabs/spaces, etc. for the project.

This avoids commits like this one, as the IDE will automatically deal with this: https://github.com/WordPress/Requests/pull/446/commits/228ec26ce827ee9081743a16fbdfc35dba45e320

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,6 @@
 codecov:
-    notify:
-        after_n_builds: 2
+  notify:
+    after_n_builds: 2
 
 coverage:
   round: nearest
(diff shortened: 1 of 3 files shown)


jrfnl (commented):
I'm happy to merge this if it helps people. Not fussed myself as my editor doesn't use the file.

Review Notes:

Not all existing (yml, json) files currently comply with these rules.

Feels to me those should be brought into sync in conjunction with this PR.

I haven't checked the PHP files, but generally speaking that's covered by the CS rules and checked anyhow.

jrfnl (approved):
:shipit:

PR #251, opened 2016-10-21
Add `Requests::has_capability()` which can be used to determine if SSL is available.

dd32:
Add `Requests::has_capability()` which can be used to determine if SSL is available.

An example of it's usage: `Requests::has_capability( array( 'ssl' => true ) );`

See #250

Issue #250, opened 2016-10-21
A Method to determine if SSL requests can be made


dd32:
WordPress <A href="https://core.trac.wordpress.org/ticket/37708">has a case</a> where we need to know if SSL requests are supported or not, currently there's no way in Requestes to know if that's possible without first attempting a SSL request.

A method to check capabilities first would be greatly appreciated!

jrfnl:
Closing as fixed via PR 492

Issue #429, opened 2020-11-10
[PHP8] curl_init can return false


TimothyBJacobs:
I don't think it's very likely, but `curl_init` can return false. `Requests_Transport_cURL` doesn't check that the return value is not false before calling `curl_setopt`. On PHP 8 this causes a fatal error.

```
Warning: Uncaught TypeError: curl_setopt(): Argument #1 ($handle) must be of type CurlHandle, bool given
```

Previously, this was just a PHP warning.

```
Warning:  curl_setopt() expects parameter 1 to be resource, boolean given
```

I think we should probably check for `$handle === false` and throw a `Requests_Exception`.

TimothyBJacobs:
This also effects `curl_multi_init` in `request_multiple`.

Dltanuub:
$curl = curl_init();

Dltanuub:
curl_setopt($curl

PR #492, opened 2021-06-18
Add Requests::has_capability() which can be used to determine if SSL is available.

jrfnl:
_Note: this is a recreation of PR #251, which was closed due to the branch rename and the remote no longer existing._
_Please see PR #251 for the original discussion on the PR._

Add `Requests::has_capability()` which can be used to determine if SSL is available.

An example of it's usage: `Requests::has_capability( array( Capability::SSL => true ) );`

See #250 and #251

/cc @dd32

---

After review of the original PR, the following additional changes have been made and are included in this commit:
* Add a `WpOrg\Requests\Capability` interface which holds constants for the queryable capabilities.
* Split the `Requests::get_transport()` method in two:
    * `Requests::get_transport_class()` which will retrieve the fully qualified class name for a working transport and will cache the result.
    * `Requests::get_transport()` to get an instantiated instance of a working transport, same as before.
* Adjusting the `Transport` interface and classes implementing it, to take a `$capabilities` parameter for the `test()` method and use the constants from the `Capability` interface internally.
* Add tests for the new `Requests::has_capabilities()` method.
* Update the `Transport` related tests and mocks.
* Add a usage example to the prose documentation.

### DIFF
--- a/docs/usage-advanced.md
+++ b/docs/usage-advanced.md
@@ -56,6 +56,16 @@ Alternatively, if you want to disable verification completely, this is possible
 with `'verify' => false`, but note that this is extremely insecure and should be
 avoided.
 
+Note that SSL verification might not be available depending on what extensions
+are enabled for your PHP environment. You can test whether Requests has
+access to a transport with SSL capabilities with the following call:
+
+```php
+use WpOrg\Requests\Capability;
+
+$ssl_available = WpOrg\Requests\Requests::test(array(Capability::SSL => true));
+```
+
 ### Security Note
 Requests supports SSL across both cURL and fsockopen in a transparent manner.
 Unlike other PHP HTTP libraries, support for verifying the certificate name is
--- a/src/Capability.php
+++ b/src/Capability.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Capability interface declaring the known capabilities.
+ *
+ * @package Requests\Utilities
+ */
+
+namespace WpOrg\Requests;
+
+/**
+ * Capability interface declaring the known capabilities.
+ *
+ * This is used as the authoritative source for which capabilities can be queried.
+ *
+ * @package Requests\Utilities
+ */
+interface Capability {
+
+	/**
+	 * Support for SSL.
+	 *
+	 * @var string
+	 */
+	const SSL = 'ssl';
+
+	/**
+	 * Collection of all capabilities supported in Requests.
+	 *
+	 * Note: this does not automatically mean that the capability will be supported for your chosen transport!
+	 *
+	 * @var array<string>
+	 */
+	const ALL = array(
+		self::SSL,
+	);
+}
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -224,47 +224,79 @@ class Requests {
 	}
 
 	/**
-	 * Get a working transport
+	 * Get the fully qualified class name (FQCN) for a working transport.
 	 *
-	 * @throws \WpOrg\Requests\Exception If no valid transport is found (`notransport`)
-	 * @return \WpOrg\Requests\Transport
+	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
+	 * @return string FQCN of the transport to use, or an empty string if no transport was
+	 *                found which provided the requested capabilities.
 	 */
-	protected static function get_transport($capabilities = array()) {
-		// Caching code, don't bother testing coverage
+	protected static function get_transport_class($capabilities = array()) {
+		// Caching code, don't bother testing coverage.
 		// @codeCoverageIgnoreStart
-		// array of capabilities as a string to be used as an array key
+		// Array of capabilities as a string to be used as an array key.
 		ksort($capabilities);
 		$cap_string = serialize($capabilities);
 
-		// Don't search for a transport if it's already been done for these $capabilities
-		if (isset(self::$transport[$cap_string]) && self::$transport[$cap_string] !== null) {
-			$class = self::$transport[$cap_string];
-			return new $class();
+		// Don't search for a transport if it's already been done for these $capabilities.
+		if (isset(self::$transport[$cap_string])) {
+			return self::$transport[$cap_string];
 		}
+
+		// Ensure we will not run this same check again later on.
+		self::$transport[$cap_string] = '';
 		// @codeCoverageIgnoreEnd
 
 		if (empty(self::$transports)) {
 			self::$transports = self::DEFAULT_TRANSPORTS;
 		}
 
-		// Find us a working transport
+		// Find us a working transport.
 		foreach (self::$transports as $class) {
 			if (!class_exists($class)) {
 				continue;
 			}
 
-			$result = call_user_func(array($class, 'test'), $capabilities);
-			if ($result) {
+			$result = $class::test($capabilities);
+			if ($result === true) {
 				self::$transport[$cap_string] = $class;
 				break;
 			}
 		}
-		if (self::$transport[$cap_string] === null) {
+
+		return self::$transport[$cap_string];
+	}
+
+	/**
+	 * Get a working transport.
+	 *
+	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
+	 * @return \WpOrg\Requests\Transport
+	 * @throws \WpOrg\Requests\Exception If no valid transport is found (`notransport`).
+	 */
+	protected static function get_transport($capabilities = array()) {
+		$class = self::get_transport_class($capabilities);
+
+		if ($class === '') {
 			throw new Exception('No working transports found', 'notransport', self::$transports);
 		}
 
-		$class = self::$transport[$cap_string];
 		return new $class();
+	}
+
+	/**
+	 * Checks to see if we have a transport for the capabilities requested.
+	 *
+	 * Supported capabilities can be found in the {@see \WpOrg\Requests\Capability}
+	 * interface as constants.
+	 *
+	 * Example usage:
+	 * `Requests::has_capabilities([Capability::SSL => true])`.
+	 *
+	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
+	 * @return bool Whether the transport has the requested capabilities.
+	 */
+	public static function has_capabilities($capabilities = array()) {
+		return self::get_transport_class($capabilities) !== '';
 	}
 
 	/**#@+
--- a/src/Transport.php
+++ b/src/Transport.php
@@ -34,8 +34,12 @@ interface Transport {
 	public function request_multiple($requests, $options);
 
 	/**
-	 * Self-test whether the transport can be used
-	 * @return bool
+	 * Self-test whether the transport can be used.
+	 *
+	 * The available capabilities to test for can be found in {@see \WpOrg\Requests\Capability}.
+	 *
+	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
+	 * @return bool Whether the transport can be used.
 	 */
-	public static function test();
+	public static function test($capabilities = array());
 }
--- a/src/Transport/Curl.php
+++ b/src/Transport/Curl.php
@@ -9,6 +9,7 @@ namespace WpOrg\Requests\Transport;
 
 use RecursiveArrayIterator;
 use RecursiveIteratorIterator;
+use WpOrg\Requests\Capability;
 use WpOrg\Requests\Exception;
 use WpOrg\Requests\Exception\InvalidArgument;
 use WpOrg\Requests\Exception\Transport\Curl as CurlException;
@@ -563,10 +564,13 @@ final class Curl implements Transport {
 	}
 
 	/**
-	 * Whether this transport is valid
+	 * Self-test whether the transport can be used.
+	 *
+	 * The available capabilities to test for can be found in {@see \WpOrg\Requests\Capability}.
 	 *
 	 * @codeCoverageIgnore
-	 * @return boolean True if the transport is valid, false otherwise.
+	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
+	 * @return bool Whether the transport can be used.
 	 */
 	public static function test($capabilities = array()) {
 		if (!function_exists('curl_init') || !function_exists('curl_exec')) {
@@ -574,7 +578,7 @@ final class Curl implements Transport {
 		}
 
 		// If needed, check that our installed curl version supports SSL
-		if (isset($capabilities['ssl']) && $capabilities['ssl']) {
+		if (isset($capabilities[Capability::SSL]) && $capabilities[Capability::SSL]) {
 			$curl_version = curl_version();
 			if (!(CURL_VERSION_SSL & $curl_version['features'])) {
 				return false;
--- a/src/Transport/Fsockopen.php
+++ b/src/Transport/Fsockopen.php
@@ -7,6 +7,7 @@
 
 namespace WpOrg\Requests\Transport;
 
+use WpOrg\Requests\Capability;
 use WpOrg\Requests\Exception;
 use WpOrg\Requests\Exception\InvalidArgument;
 use WpOrg\Requests\Port;
@@ -448,10 +449,13 @@ final class Fsockopen implements Transport {
 	}
 
 	/**
-	 * Whether this transport is valid
+	 * Self-test whether the transport can be used.
+	 *
+	 * The available capabilities to test for can be found in {@see \WpOrg\Requests\Capability}.
 	 *
 	 * @codeCoverageIgnore
-	 * @return boolean True if the transport is valid, false otherwise.
+	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
+	 * @return bool Whether the transport can be used.
 	 */
 	public static function test($capabilities = array()) {
 		if (!function_exists('fsockopen')) {
@@ -459,7 +463,7 @@ final class Fsockopen implements Transport {
 		}
 
 		// If needed, check that streams support SSL
-		if (isset($capabilities['ssl']) && $capabilities['ssl']) {
+		if (isset($capabilities[Capability::SSL]) && $capabilities[Capability::SSL]) {
 			if (!extension_loaded('openssl') || !function_exists('openssl_x509_parse')) {
 				return false;
 			}
--- a/tests/RequestsTest.php
+++ b/tests/RequestsTest.php
@@ -2,12 +2,14 @@
 
 namespace WpOrg\Requests\Tests;
 
+use ReflectionProperty;
+use WpOrg\Requests\Capability;
 use WpOrg\Requests\Exception;
 use WpOrg\Requests\Requests;
 use WpOrg\Requests\Response\Headers;
 use WpOrg\Requests\Tests\Fixtures\RawTransportMock;
+use WpOrg\Requests\Tests\Fixtures\TestTransportMock;
 use WpOrg\Requests\Tests\Fixtures\TransportMock;
-use WpOrg\Requests\Tests\TestCase;
 
 final class RequestsTest extends TestCase {
 
@@ -168,5 +170,31 @@ final class RequestsTest extends TestCase {
 		$this->expectException(Exception::class);
 		$this->expectExceptionMessage('timed out');
 		Requests::get(httpbin('/delay/3'), array(), $options);
+	}
+
+	/**
+	 * @covers \WpOrg\Requests\Requests::has_capabilities
+	 */
+	public function testHasCapabilitiesSucceedsForDetectingSsl() {
+		if (!extension_loaded('curl') && !extension_loaded('openssl')) {
+			$this->markTestSkipped('Testing for SSL requires either the curl or the openssl extension');
+		}
+		$this->assertTrue(Requests::has_capabilities(array(Capability::SSL => true)));
+	}
+
+	/**
+	 * @covers \WpOrg\Requests\Requests::has_capabilities
+	 */
+	public function testHasCapabilitiesFailsForUnsupportedCapabilities() {
+		$transports = new ReflectionProperty(Requests::class, 'transports');
+		$transports->setAccessible(true);
+		$transports->setValue(array(TestTransportMock::class));
+
+		$result = Requests::has_capabilities(array('time-travel' => true));
+
+		$transports->setValue(array());
+		$transports->setAccessible(false);
+
+		$this->assertFalse($result);
 	}
 }
--- a/tests/Transport/BaseTestCase.php
+++ b/tests/Transport/BaseTestCase.php
@@ -3,6 +3,7 @@
 namespace WpOrg\Requests\Tests\Transport;
 
 use stdClass;
+use WpOrg\Requests\Capability;
 use WpOrg\Requests\Exception;
 use WpOrg\Requests\Exception\Http\StatusUnknown;
 use WpOrg\Requests\Exception\InvalidArgument;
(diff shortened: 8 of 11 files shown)


schlessera:
Having a way to test for a capability makes sense, but the `$capabilities` array is not the best way to define what to actually test.

We'll rework this in the following way:
- Change the signature to `has_capability( string $capability )`.
- Create an interface `Capability` which contains constants for all the known capabilities (which seems to be only `ssl` right now). This will mean the usage looks like this: `Requests::has_capability( Capability::SSL )`.
- Document usage for `has_capability()` on where to find the list of supported capabilities.
- Add tests.

schlessera:
While going through the code to make the changes I suggested above, I noticed an issue with the current implementation.

The thing is that the way this is implemented, it always triggers the transport constructor if it has the capabilities, only to then discard that transport (the check returns a boolean only).

While the `Requests_Transport_fsockopen` doesn't have a constructor at all, the `Requests_Transport_cURL` (which should be used in most cases) has a constructor that configure cURL and then create a new resource handle, only to then discard that resource handle again in the destructor.

I think this is extremely wasteful for such a simple check. Avoiding this will need a change to the `Requests::get_transport()` method as well.

schlessera:
@jrfnl I can't set you as reviewer, as you've created the PR. Please review this latest change.

tests/Mock/TestTransportMock.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Requests\Tests\Mock;
+
+use Requests_Transport;
+
+class TestTransportMock implements Requests_Transport {
```
jrfnl:
```suggestion
class TransportTestMock implements Requests_Transport {
```
(class/interface name - method name - mock)

library/Requests/Capability.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Capability interface declaring the known capabilities
+ *
+ * This is used as discovery to figure out which capabilities can be queried.
+ *
+ * @package Requests
+ * @subpackage Capability
+ */
+class Requests_Capability {
```
jrfnl: Should this be a class or an interface ?

If a **class**:
* Please update the file and class docblock to remove the confusion.
* Should there be a `has()` method with a signature along the lines of: `has(Transport $transport, $capability) : bool`, which calls the `Transport::test()` method ? Or maybe even, should (a variation of) the new `Requests::get_transport_class()` method be moved here ?

If an **interface**:
* Should the interface extend the `Transport` interface ?
* Should the `test` method stub be moved from the `Transport` interface to the `Capability` interface ?

library/Requests/Transport/cURL.php
```diff
@@ -575,7 +575,7 @@ public static function test($capabilities = array()) {
 		}
 
 		// If needed, check that our installed curl version supports SSL
-		if (isset($capabilities['ssl']) && $capabilities['ssl']) {
+		if (isset($capabilities[Requests_Capability::SSL]) && $capabilities[Requests_Capability::SSL]) {
```
jrfnl: Might be helpful to update the docblock of the method to actually mention and describe the format of the `$capabilities` array parameter and which capabilities are (currently) supported to test for.

library/Requests/Transport/fsockopen.php
```diff
@@ -459,7 +459,7 @@ public static function test($capabilities = array()) {
 		}
 
 		// If needed, check that streams support SSL
-		if (isset($capabilities['ssl']) && $capabilities['ssl']) {
+		if (isset($capabilities[Requests_Capability::SSL]) && $capabilities[Requests_Capability::SSL]) {
```
jrfnl: Might be helpful to update the docblock of the method to actually mention and describe the format of the `$capabilities` array parameter and which capabilities are (currently) supported to test for.

tests/RequestsTest.php
```diff
@@ -169,4 +172,19 @@ public function testTimeoutException() {
 		$this->expectExceptionMessage('timed out');
 		Requests::get(httpbin('/delay/3'), array(), $options);
 	}
+
+	public function testHasCapabilitiesSucceedsForDetectingSsl() {
```
jrfnl, line 178: Does this test need an `@requires openssl` tag ?

Also, please add a `@covers` tag for the test (and maybe a short description ?)

tests/Transport/BaseTestCase.php
```diff
@@ -21,7 +22,7 @@ public function set_up() {
 			return;
 		}
 
-		$ssl_supported = call_user_func($callback, array('ssl' => true));
+		$ssl_supported = call_user_func($callback, array(Requests_Capability::SSL => true));
```
jrfnl: Should we modernize this `set_up()` method, while we're at it, by using `$transport::test()` directly instead of using `call_user_func()` ? (here and on line 18)

(Can be left for the test review to keep this PR atomic)

tests/RequestsTest.php
```diff
@@ -169,4 +172,19 @@ public function testTimeoutException() {
...
+		$this->assertTrue(Requests::has_capabilities(array(Requests_Capability::SSL => true)));
+	}
+
+	public function testHasCapabilitiesFailsForUnsupportedCapabilities() {
```
jrfnl, line 188: Please add a `@covers` tag for the test (and maybe a short description ?)

tests/RequestsTest.php
```diff
@@ -169,4 +172,19 @@ public function testTimeoutException() {
...
+		$transports = new ReflectionProperty(Requests::class, 'transports');
+		$transports->setAccessible(true);
+		$transports->setValue(array(TestTransportMock::class));
```
jrfnl, line 191: Why are you using Reflection here instead of calling `Requests::add_transport()` ?

And on that note: should there be a `Requests::remove_transport()` method ?

tests/RequestsTest.php
```diff
@@ -169,4 +172,19 @@ public function testTimeoutException() {
...
+
+		$this->assertFalse(Requests::has_capabilities(array('time-travel' => true)));
+
+		$transports->setValue(null);
+		$transports->setAccessible(false);
```
jrfnl: Please do the "undoing" before the assertion as otherwise the class may be left in an unstable state for follow-on tests if the test would fail.

```php
$result = Requests::has_capabilities(array('time-travel' => true));

$transports->setValue(null);
$transports->setAccessible(false);

$this->assertFalse($result);
```

library/Requests.php
```diff
@@ -205,20 +204,47 @@ protected static function get_transport($capabilities = array()) {
...
 
-		$class = self::$transport[$cap_string];
 		return new $class();
 	}
 
+	/**
+	 * Checks to see if we have a transport for the capabilities requested
+	 *
+	 * Supported capabilities can be found in the {@see Requests_Capability}
+	 * interface as constants.
```
jrfnl:
```suggestion
	 * class as constants.
```

`interface` or `class` ? (see my other remark about this elsewhere)

library/Requests/Capability.php
```diff
@@ -0,0 +1,34 @@
...
+
+	/**
+	 * Support for SSL.
+	 *
+	 * @var string
+	 */
+	const SSL = 'ssl';
+
+	/**
+	 * Collection of all known capabilities.
```
jrfnl: Looks like this constant isn't used. Do we want to add dead code ?

```suggestion
	 * Collection of all capabilities supported in Requests.
	 *
	 * Note: this does not automatically mean that the capability will be supported for your chosen transport!
```

library/Requests.php
```diff
@@ -173,12 +173,12 @@ public static function add_transport($transport) {
 	}
 
 	/**
-	 * Get a working transport
+	 * Get the FQCN for a working transport
 	 *
-	 * @throws Requests_Exception If no valid transport is found (`notransport`)
-	 * @return Requests_Transport
+	 * @return string FQCN of the transport to use, or an empty string if none
+	 *                provided the requested capabilities.
```
jrfnl:
```suggestion
	 * @return string FQCN of the transport to use, or an empty string if no transport was
	 *                found which provided the requested capabilities.
```

library/Requests.php
```diff
@@ -205,20 +204,47 @@ protected static function get_transport($capabilities = array()) {
...
+
+		return '';
+	}
+
+	/**
+	 * Get a working transport
+	 *
+	 * @throws Requests_Exception If no valid transport is found (`notransport`)
+	 * @return Requests_Transport
+	 */
```
jrfnl: Please add a `@param` tag to the docblock and describe the format of the `$capabilities` array parameter.

library/Requests.php
```diff
@@ -173,12 +173,12 @@ public static function add_transport($transport) {
...
 
 	/**
-	 * Get a working transport
+	 * Get the FQCN for a working transport
 	 *
-	 * @throws Requests_Exception If no valid transport is found (`notransport`)
-	 * @return Requests_Transport
+	 * @return string FQCN of the transport to use, or an empty string if none
+	 *                provided the requested capabilities.
 	 */
```
jrfnl: Please add a `@param` tag to the docblock and describe the format of the `$capabilities` array parameter.

library/Requests.php
```diff
@@ -187,8 +187,7 @@ protected static function get_transport($capabilities = array()) {
 
 		// Don't search for a transport if it's already been done for these $capabilities
 		if (isset(self::$transport[$cap_string]) && self::$transport[$cap_string] !== null) {
-			$class = self::$transport[$cap_string];
-			return new $class();
+			return self::$transport[$cap_string];
 		}
 		// @codeCoverageIgnoreEnd
```
jrfnl: 👇🏻 I'm noticing that the double listing of the "default" transports here and in the above `add_transport()` method is bugging me. Will add a fix for this to the "magic constants" set of commits.

library/Requests.php
```diff
@@ -205,20 +204,47 @@ protected static function get_transport($capabilities = array()) {
...
+	/**
+	 * Get a working transport
+	 *
+	 * @throws Requests_Exception If no valid transport is found (`notransport`)
+	 * @return Requests_Transport
+	 */
+	protected static function get_transport($capabilities = array()) {
+		$class = self::get_transport_class($capabilities);
+
+		if ($class === '') {
```
jrfnl: Is the check against an empty string sufficient here ? Or should this check that the `$class` is not listed in the array of valid transports ? (code hardening; check for what you want, not for what you don't want)

jrfnl (commented):
Gone through it in detail and left some comments and questions. I look forward to the next iteration.

jrfnl:
Oh and one more question: should this method and this new class be mentioned anywhere in the docs/are there docs which need updating for the changes which will now be made ?

tests/Mock/TestTransportMock.php
schlessera: It's supposed to be a type of transport, not a method name.
We already have these two in the code:
- `RawTransportMock`
- `TransportMock`

library/Requests/Capability.php
schlessera: > Should this be a class or an interface ?

Ideally, it'd be an enum. However, given there's no clean way to do that in PHP 5.6 (at least without additional dependencies), I went with an interface to just keep track of the constants for now. I'll adapt the definition.

> Should the interface extend the Transport interface ?

No, it is not of type "transport", it is just a collection of constants (for now).

> Should the test method stub be moved from the Transport interface to the Capability interface ?

The `test()` method actually does multiple things and is not exclusively tied to testing capabilities, therefore I'd keep it as-is for now. If we ever want to make this cleaner, it would require a more involved refactor.

tests/RequestsTest.php
schlessera: > Why are you using Reflection here instead of calling Requests::add_transport() ?

I'm using reflection because there is no way of replacing the existing transports, you can only add to them. And this specific test requires the capabilities call to fail on a certain condition, which cannot be done as soon as one of the already existing transports is registered.

> And on that note: should there be a Requests::remove_transport() method ?

Something seems to be missing for more control yes. Not sure removal is the best approach, though, maybe filtering in some form would be preferable.

library/Requests.php
schlessera: We don't have a list of valid transports that includes possible extensions.

The empty string is the "error case" that is being returned by `get_transport_class()`, I used an empty string so that I can keep a strong type of `string`.

schlessera:
@jrfnl This can be reviewed again.

jrfnl:
> The thing is that the way this is implemented, it always triggers the transport constructor if it has the capabilities, only to then discard that transport (the check returns a boolean only).
> 
> While the `Requests_Transport_fsockopen` doesn't have a constructor at all, the `Requests_Transport_cURL` (which should be used in most cases) has a constructor that configure cURL and then create a new resource handle, only to then discard that resource handle again in the destructor.
> 
> I think this is extremely wasteful for such a simple check. Avoiding this will need a change to the `Requests::get_transport()` method as well.

There may actually be good reason for this - I think this may be loosely related to #429.

schlessera:
@jrfnl This should be ready for a new review.

docs/usage-advanced.md
```diff
@@ -56,6 +56,14 @@ Alternatively, if you want to disable verification completely, this is possible
 with `'verify' => false`, but note that this is extremely insecure and should be
 avoided.
 
+Note that SSL verification might not be available depending on what extensions
+are being loaded for your PHP environment. You can test whether Requests has
+access to a transport with SSL capabilities with the following call:
```
jrfnl:
```suggestion
Note that SSL verification might not be available depending on what extensions
are enabled for your PHP environment. You can test whether Requests has
access to a transport with SSL capabilities with the following call:
```

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
 	}
 
 	/**
-	 * Get a working transport
+	 * Get the FQCN for a working transport.
```
jrfnl:
```suggestion
	 * Get the fully qualified class name (FQCN) for a working transport.
```

(Lower the threshold for entry for less experienced contributors and users of the API docs)

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
...
 	 *
-	 * @throws \WpOrg\Requests\Exception If no valid transport is found (`notransport`)
-	 * @return \WpOrg\Requests\Transport
+	 * @param array<string, bool> $capabilities Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```
jrfnl:
```suggestion
	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
...
 		}
-		if (self::$transport[$cap_string] === null) {
+
+		return '';
+	}
+
+	/**
+	 * Get a working transport.
+	 *
+	 * @param array<string, bool> $capabilities Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```
jrfnl:
```suggestion
	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
...
+		// Caching code, don't bother testing coverage.
 		// @codeCoverageIgnoreStart
-		// array of capabilities as a string to be used as an array key
+		// Array of capabilities as a string to be used as an array key.
 		ksort($capabilities);
 		$cap_string = serialize($capabilities);
 
-		// Don't search for a transport if it's already been done for these $capabilities
+		// Don't search for a transport if it's already been done for these $capabilities.
 		if (isset(self::$transport[$cap_string]) && self::$transport[$cap_string] !== null) {
```
jrfnl:
```php
self::$transport[$cap_string] !== null
```

Is it (still) possible for this condition to ever be `false` ?

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
...
-		// Find us a working transport
+		// Find us a working transport.
 		foreach (self::$transports as $class) {
 			if (!class_exists($class)) {
 				continue;
 			}
 
-			$result = call_user_func(array($class, 'test'), $capabilities);
+			$result = $class::test($capabilities);
 			if ($result) {
```
jrfnl: Let's make this check type specific ?
```suggestion
			if ($result === true) {
```

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
...
 				self::$transport[$cap_string] = $class;
-				break;
+				return $class;
 			}
 		}
-		if (self::$transport[$cap_string] === null) {
+
```
jrfnl, line 265: The caching is now broken. When no transport is found for this particular `$cap_string`, a cache entry should still be set (to empty string), otherwise the whole check will be done time and time again.

src/Requests.php
```diff
@@ -224,49 +224,78 @@ public static function add_transport($transport) {
...
+	/**
+	 * Checks to see if we have a transport for the capabilities requested.
+	 *
+	 * Supported capabilities can be found in the {@see \WpOrg\Requests\Capability}
+	 * interface as constants.
+	 *
+	 * Example usage:
+	 * `Requests::has_capabilities([Capability::SSL => true])`.
+	 *
+	 * @param array<string, bool> $capabilities Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```
jrfnl:
```suggestion
	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```

src/Transport.php
```diff
@@ -36,8 +36,12 @@ public function request($url, $headers = array(), $data = array(), $options = ar
 	public function request_multiple($requests, $options);
 
 	/**
-	 * Self-test whether the transport can be used
-	 * @return bool
+	 * Self-test whether the transport can be used.
+	 *
+	 * The available capabilities to test for can be found in {@see \WpOrg\Requests\Capability}.
+	 *
+	 * @param array<string, bool> $capabilities Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```
jrfnl:
```suggestion
	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```

src/Transport/Curl.php
```diff
@@ -565,18 +566,21 @@ private static function format_get($url, $data) {
...
 
 	/**
-	 * Whether this transport is valid
+	 * Self-test whether the transport can be used.
+	 *
+	 * The available capabilities to test for can be found in {@see \WpOrg\Requests\Capability}.
 	 *
 	 * @codeCoverageIgnore
-	 * @return boolean True if the transport is valid, false otherwise.
+	 * @param array<string, bool> $capabilities Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```
```suggestion
	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```

src/Transport/Fsockopen.php
```diff
@@ -450,18 +451,21 @@ public function verify_certificate_from_context($host, $context) {
...
 
 	/**
-	 * Whether this transport is valid
+	 * Self-test whether the transport can be used.
+	 *
+	 * The available capabilities to test for can be found in {@see \WpOrg\Requests\Capability}.
 	 *
 	 * @codeCoverageIgnore
-	 * @return boolean True if the transport is valid, false otherwise.
+	 * @param array<string, bool> $capabilities Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```
jrfnl:
```suggestion
	 * @param array<string, bool> $capabilities Optional. Associative array of capabilities to test against, i.e. `['<capability>' => true]`.
```

src/Capability.php
```diff
@@ -0,0 +1,38 @@
...
+ * @package Requests
+ * @subpackage Capability
+ */
+
+namespace WpOrg\Requests;
+
+/**
+ * Capability interface declaring the known capabilities.
+ *
+ * This is used as discovery to figure out which capabilities can be queried.
```
jrfnl:
```suggestion
 * This is used as the authoritative source for which capabilities can be queried.
```

src/Capability.php
```diff
@@ -0,0 +1,38 @@
...
+
+	/**
+	 * Collection of all capabilities supported in Requests.
+	 *
+	 * Note: this does not automatically mean that the capability will be supported for your chosen transport!
+	 *
+	 * @var array<string>
+	 */
+	const ALL = array(
+		self::SSL,
```
jrfnl, line 34: Would it be an idea to have this set up as an array like so ? That way `isset()` can also be used, which could proof more performant in certain situations (thinking ahead to future uses of the array).
```php
    self::SSL => self::SSL,
```

src/Requests.php
jrfnl: Should this comment contain information on how **not** setting the property will affect the function result ?
I.e. something along the lines of "_If passed, a transport will only be considered valid if it also supports the requested capabilities. If the parameter is not passed, a transport will be considered as valid if available, independently of its capabilities._"

(here and everywhere else)

tests/RequestsTest.php
```diff
@@ -169,4 +171,30 @@ public function testTimeoutException() {
...
+	 * @covers \WpOrg\Requests\Requests::has_capabilities
+	 */
+	public function testHasCapabilitiesFailsForUnsupportedCapabilities() {
+		$transports = new ReflectionProperty(Requests::class, 'transports');
+		$transports->setAccessible(true);
+		$transports->setValue(array(TestTransportMock::class));
+
+		$result = Requests::has_capabilities(array('time-travel' => true));
+
+		$transports->setValue(null);
```
jrfnl:
```suggestion
		$transports->setValue(array());
```

The default value in the `Requests` class is an empty array, so let's revert back to that instead of introducing a new variable type (`null`) which may cause side effects.

https://github.com/WordPress/Requests/blob/afa69a60b32c5b9263266711e5f635938c2fc39f/src/Requests.php#L175

tests/Transport/BaseTestCase.php
```diff
@@ -17,15 +18,17 @@ abstract class BaseTestCase extends TestCase {
 	protected $skip_https = false;
 
 	public function set_up() {
-		$callback  = array($this->transport, 'test');
-		$supported = call_user_func($callback);
+		// Intermediary variable $test_method can be skipped with PHP 7+.
```
jrfnl:
```suggestion
		// Intermediary variable $test_method. This can be simplified (removed) once the minimum supported PHP version is 7.0 or higher.
```

jrfnl (commented):
@schlessera Gone through it with a fine toothcomb. Hope I caught everything. Nearly there! Thanks for all your work on this.

src/Capability.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/**
+ * Capability interface declaring the known capabilities.
+ *
+ * @package Requests
+ * @subpackage Capability
```
jrfnl: Related to #590 - should this be changed ? Do we want to add a new subcategory or does this belong in `Utilities` ?

```suggestion
 * @package Requests\Utilities
```

tests/RequestsTest.php
schlessera: The test still runs without error without the OpenSSL extension, so I'd say the `@requires` annotation is not needed (at least not in this state).

src/Capability.php
schlessera: This is part of the public interface, so I think the regular array will convey a clearer meaning. Also, I don't see this array growing to a large size, so I'm not sure the performance difference between hashed and non-hashed lookup is worth doubling the memory consumption in this particular case.

src/Requests.php
schlessera: I think the caching never worked in the first place. The original code already stored something when the test produced a positive result...

I'll fix it anyway, as I think the intention clearly was to use caching.

src/Capability.php
jrfnl: Fair. We can always adjust, if needed, later.

docs/usage-advanced.md
```diff
@@ -56,6 +56,14 @@ Alternatively, if you want to disable verification completely, this is possible
 with `'verify' => false`, but note that this is extremely insecure and should be
 avoided.
 
+Note that SSL verification might not be available depending on what extensions
+are enabled for your PHP environment. You can test whether Requests has
+access to a transport with SSL capabilities with the following call:
+
+```php
+$ssl_available = WpOrg\Requests\Requests::test(array('ssl' => true));
```
jrfnl:
```suggestion
use WpOrg\Requests\Capability;

$ssl_available = WpOrg\Requests\Requests::test(array(Capability::SSL => true));
```

src/Capability.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace WpOrg\Requests;
+
+/**
+ * Capability interface declaring the known capabilities.
+ *
+ * This is used as the authoritative source for which capabilities can be queried.
+ *
+ * @package Requests
+ * @subpackage Capability
```
jrfnl:
```suggestion
 * @package Requests\Utilities
```

src/Requests.php
jrfnl: It wasn't broken before - the `$class = self::$transport[$cap_string];` (old line 266) in the original method handled this, but that line was removed when the method was split into two.

jrfnl (commented):
Nearly done, will fix up those last few things now.

jrfnl:
Ready for merge IMO

schlessera (approved):

PR #494, opened 2021-06-18
max_bytes not respected in a performant manner

jrfnl:
_Note: this is a recreation of PR #337, which was closed due to the branch rename and the remote no longer existing._
_Please see PR #337 for the original discussion on the PR._

With some testing I was noticing that a request for 100 byes from a 10MB file was timing out after 5 seconds.

Turns out that the cause was that Requests downloads the full 10MB file in order to return the first 100 bytes of it.

This PR simply aborts the connection when enough bytes have been read from the stream.

/cc @dd32

### DIFF
--- a/src/Transport/Curl.php
+++ b/src/Transport/Curl.php
@@ -210,7 +210,7 @@ final class Curl implements Transport {
 
 		if ($curl_errno === CURLE_WRITE_ERROR
 			&& $this->response_byte_limit
-			&& $this->response_bytes === $this->response_byte_limit
+			&& $this->response_bytes >= $this->response_byte_limit
 		) {
 			// Not actually an error in this case. We've drained all the data from the request that we want.
 			$curl_errno = false;
@@ -497,7 +497,7 @@ final class Curl implements Transport {
 		$curl_errno = curl_errno($this->handle);
 		if ($curl_errno === CURLE_WRITE_ERROR
 			&& $this->response_byte_limit
-			&& $this->response_bytes === $this->response_byte_limit
+			&& $this->response_bytes >= $this->response_byte_limit
 		) {
 			// Not actually an error in this case. We've drained all the data from the request that we want.
 			$curl_errno = false;
@@ -566,7 +566,9 @@ final class Curl implements Transport {
 		}
 
 		if ($this->stream_handle) {
-			fwrite($this->stream_handle, $data);
+			if ($data !== '') {
+				fwrite($this->stream_handle, $data);
+			}
 		} else {
 			$this->response_data .= $data;
 		}
--- a/src/Transport/Curl.php
+++ b/src/Transport/Curl.php
@@ -208,8 +208,11 @@ final class Curl implements Transport {
 
 		$curl_errno = curl_errno($this->handle);
 
-		if ($curl_errno === CURLE_WRITE_ERROR && $this->response_byte_limit && $this->response_byte_limit === $this->response_bytes) {
-			// Not actually an error in this case. We've drained as much data from the request that we want.
+		if ($curl_errno === CURLE_WRITE_ERROR
+			&& $this->response_byte_limit
+			&& $this->response_bytes === $this->response_byte_limit
+		) {
+			// Not actually an error in this case. We've drained all the data from the request that we want.
 			$curl_errno = false;
 		}
 
@@ -492,8 +495,11 @@ final class Curl implements Transport {
 		}
 
 		$curl_errno = curl_errno($this->handle);
-		if ($curl_errno === CURLE_WRITE_ERROR && $this->response_byte_limit && $this->response_byte_limit === $this->response_bytes) {
-			// Not actually an error in this case. We've drained as much data from the request that we want.
+		if ($curl_errno === CURLE_WRITE_ERROR
+			&& $this->response_byte_limit
+			&& $this->response_bytes === $this->response_byte_limit
+		) {
+			// Not actually an error in this case. We've drained all the data from the request that we want.
 			$curl_errno = false;
 		}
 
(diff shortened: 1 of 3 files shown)


schlessera:
Updates needed for this PR to be mergeable:
- Tests need to be added.
- PHP constants should be used for the cURL error codes, instead of magic numbers (see [licurl error codes](https://curl.se/libcurl/c/libcurl-errors.html) & [curl constants](https://www.php.net/manual/en/curl.constants.php)).

jrfnl:
>     * PHP constants should be used for the cURL error codes, instead of magic numbers (see [licurl error codes](https://curl.se/libcurl/c/libcurl-errors.html) & [curl constants](https://www.php.net/manual/en/curl.constants.php)).

This point has been addressed in the recent rebase to solve merge conflicts.

This PR does however still needs tests.

schlessera:
@jrfnl I've added tests for this. The tests could verify the problem, i.e. the tests failed without the fix, and they now succeed with the fix. I've drastically increased the amount of bytes being requested by the server to force the issue, but with the fix in place, the download should still only be the first few bytes anyway.

tests/Transport/BaseTestCase.php
```diff
@@ -1172,4 +1174,29 @@ public function testBodyDataFormat() {
...
+			function ($block, $size, $max_bytes) {
+				static $passed_max_bytes = false;
+				$this->assertFalse(
+					$passed_max_bytes,
+					'Failed asserting that body download was stopped as soon as max_bytes was reached.'
+				);
+				if ($size >= $max_bytes) {
+					$passed_max_bytes = true;
+				}
+			}
```
jrfnl: I must be missing something...

The way this looks to me, the assertion is checking the state of the _previous_ call to this callback, not the current call. Is that correct and if so, why ?

I have a feeling this could this be simplified to:
```php
				$this->assertLessThan(
					$max_bytes,
					$size,
					'Failed asserting that body download was stopped as soon as max_bytes was reached.'
				);
```

tests/Transport/BaseTestCase.php
```diff
@@ -49,18 +49,20 @@ public function testResponseByteLimit() {
 		$limit    = 104;
 		$options  = [
 			'max_bytes' => $limit,
+			'hooks'     => $this->getMaxBytesAssertionHooks(),
 		];
-		$response = Requests::get(httpbin('/bytes/325'), [], $this->getOptions($options));
+		$response = Requests::get(httpbin('/bytes/1000000'), [], $this->getOptions($options));
```
jrfnl: Just to understand and document why: why does this need to be such a high number ? What are the factors influencing this ?

FYI: I can reproduce the failure without the fix and pass with the fix, with as little as 2500 bytes (locally on Windows 10).

tests/Transport/BaseTestCase.php
```diff
@@ -49,18 +49,20 @@ public function testResponseByteLimit() {
 		$limit    = 104;
```
jrfnl, line 55: On a loosely (un)related note: this is one of the tests which always fails on my Windows machine locally:
```
1) WpOrg\Requests\Tests\Transport\FsockopenTest::testResponseByteLimit
Failed asserting that 99 is identical to 104.

path/to/Requests/tests/Transport/BaseTestCase.php:55
```

jrfnl:
Rebased without changes to get it past an imaginary merge conflict.

src/Transport/Curl.php
```diff
@@ -206,7 +206,14 @@ public function request($url, $headers = [], $data = [], $options = []) {
 
 		$options['hooks']->dispatch('curl.after_send', []);
 
-		if (curl_errno($this->handle) === CURLE_WRITE_ERROR || curl_errno($this->handle) === CURLE_BAD_CONTENT_ENCODING) {
+		$curl_errno = curl_errno($this->handle);
+
+		if ($curl_errno === CURLE_WRITE_ERROR && $this->response_byte_limit && $this->response_byte_limit === $this->response_bytes) {
```
```suggestion
		if ($curl_errno === CURLE_WRITE_ERROR && $this->response_byte_limit && $this->response_bytes >= $this->response_byte_limit) {
```

Just to be on the safe side ?

src/Transport/Curl.php
```diff
@@ -481,7 +488,13 @@ public function process_response($response, $options) {
 			$this->headers .= $response;
 		}
 
-		if (curl_errno($this->handle)) {
+		$curl_errno = curl_errno($this->handle);
+		if ($curl_errno === CURLE_WRITE_ERROR && $this->response_byte_limit && $this->response_byte_limit === $this->response_bytes) {
```
```suggestion
		if ($curl_errno === CURLE_WRITE_ERROR && $this->response_byte_limit && $this->response_byte_limit >= $this->response_bytes) {
```

src/Transport/Curl.php
```diff
@@ -536,15 +549,10 @@ public function stream_body($handle, $data) {
 
 		// Are we limiting the response size?
 		if ($this->response_byte_limit) {
-			if ($this->response_bytes === $this->response_byte_limit) {
-				// Already at maximum, move on
-				return $data_length;
-			}
-
```
jrfnl, line 546: As discussed, I think it would be good to bring back this condition and change the `===` to `>=` as this is a `public` method.

jrfnl:
Rebased without changes to get it past an imaginary merge conflict.

jrfnl (commented):
Just some review notes

tests/Transport/BaseTestCase.php
schlessera: I used high numbers because curl might decide to use bigger chunks, and I wanted to ensure that we can reliably say that not only a single chunk going past the limit is truncated, but also that all subsequent chunks have been suppressed.

src/Transport/Curl.php
schlessera: As per our discussion: there should be a check to awoid a zero byte `fwrite()`, but any check needs to be after the calculations of the chunk sizes, not before.

schlessera (approved):

PR #499, opened 2021-06-21
PHP 8.1: improve input validation for `Requests_Transport_(fsockopen|cURL)`

jrfnl:
PHP 8.1 deprecates passing `null` to non-nullable parameters for PHP native functions.
Ref: https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg

The `Requests_Transport_fsockopen::request()` method passes `$request_body` into the PHP native `strlen()` function without sufficient validation, which on PHP 8.1 results in a `strlen(): Passing null to parameter #1 ($string) of type string is deprecated` notification.

The existing `RequestsTest_Transport_Base::testEmptyPOST()` test method exposed this.

When investigating this issue, I realized that no significant input validation was being done on the `$data` parameter.
PR #368 in response to issue #248 added a test to verify that the "content-length" header was correctly set when `null` would be passed as `$data` and added a safe-guard specifically for when `null` would be passed. Also, as a matter of form, integers and floats were handled correctly for `fsockopen`, but anything else would lead to PHP errors in unexpected places in every supported PHP version, including integers and floats in combination with the `cURL` class.

To mitigate this, I propose to:
* Add a `Requests_Exception_InvalidArgument` class which extends the PHP native `InvalidArgumentException`.
* Add input validation for the `$data` parameter to both the `cURL` as well as the `fsockopen` Transport class which maintains and stabilizes the pre-existing behaviour for handling of `null`, synchronizes the behaviour for integers and floats to be the same across `fsockopen` and `cURL`, but will throw the new `InvalidArgument` exception for any other type of unexpected input for the `$data` parameter.

Includes additional tests covering this change.

### DIFF
--- a/library/Requests/Transport/cURL.php
+++ b/library/Requests/Transport/cURL.php
@@ -130,6 +130,25 @@ class Requests_Transport_cURL implements Requests_Transport {
 	 * @return string Raw HTTP result
 	 */
 	public function request($url, $headers = array(), $data = array(), $options = array()) {
+		if (!is_array($data) && !is_string($data)) {
+			if ($data === null) {
+				$data = '';
+			} elseif (is_int($data) || is_float($data)) {
+				$data = (string) $data;
+			} else {
+				throw new Requests_Exception_InvalidArgument(
+					sprintf(
+						'%s: Argument #%d (%s) must be of type %s, %s given',
+						__METHOD__,
+						3,
+						'$data',
+						'array|string',
+						gettype($data)
+					)
+				);
+			}
+		}
+
 		$this->hooks = $options['hooks'];
 
 		$this->setup_handle($url, $headers, $data, $options);
--- a/library/Requests/Transport/fsockopen.php
+++ b/library/Requests/Transport/fsockopen.php
@@ -56,6 +56,25 @@ class Requests_Transport_fsockopen implements Requests_Transport {
 	 * @return string Raw HTTP result
 	 */
 	public function request($url, $headers = array(), $data = array(), $options = array()) {
+		if (!is_array($data) && !is_string($data)) {
+			if ($data === null) {
+				$data = '';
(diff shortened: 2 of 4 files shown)


library/Requests/Transport/cURL.php
```diff
@@ -130,6 +130,25 @@ public function __destruct() {
 	 * @return string Raw HTTP result
 	 */
 	public function request($url, $headers = array(), $data = array(), $options = array()) {
+		if (is_array($data) === false && is_string($data) === false) {
```
schlessera: I think this type of notation might be easier to read:
```suggestion
		if (! is_array($data) && ! is_string($data)) {
```
Is there a specific reason to use the `=== false` notation instead?

library/Requests/Transport/fsockopen.php
```diff
@@ -56,6 +56,25 @@ class Requests_Transport_fsockopen implements Requests_Transport {
...
+						'%s: Argument #%d (%s) must be of type %s, %s given',
+						__METHOD__,
+						3,
+						'$data',
+						'array|string',
+						gettype($data)
+					)
+				);
+			}
+		}
```
schlessera: As the entire check is duplicated, I suggest we extract this into a protected method of the base transport class:
```suggestion
		$data = $this->validate_data_tyoe($data);
```

schlessera (changes requested):

library/Requests/Transport/fsockopen.php
schlessera: We don't have a base class, only an interface. So we'll skip this refactor for now.

library/Requests/Transport/cURL.php
jrfnl: Fixed

schlessera (approved):

PR #520, opened 2021-08-30
FsockopenTest: add regression test for HTTP string

jrfnl:
PR #339 made a change to safeguard that the HTTP string would always contain a decimal point, never a comma.

The original bug was related to the current locale influencing the results of the call to `sprintf()`.

At the time, no test was added to safeguard the fix against regressions.

This PR now adds this regression test.

I've verified that without the fix as merged in #339, this test would fail.

Includes adjusting the GH Actions workflows to install extra locales as otherwise the test would fail on the locale being changed to in the test not being available.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -79,6 +79,9 @@ jobs:
       - name: Access localhost on port 9002
         run: curl -i http://localhost:9002
 
+      - name: Show available locales
+        run: locale -a
+
       - name: Show PHPUnit version
         run: vendor/bin/phpunit --version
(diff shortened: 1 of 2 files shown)


jrfnl:
Closing for the time being while debugging why the locale change is not working on GHA.

jrfnl:
The locale change needed a tweak to the GH Actions workflows. Re-opening the PR now this has been fixed.

jrfnl:
Oh .. great.. can't reopen once there's been a force-push. Oh well. Opening a new PR instead in that case. See #521

PR #527, opened 2021-09-05
document how to set a response timeout

shmooth:
i'm not sure where i found this info, but it was one of the first things i went looking for, and couldn't find it in the docs, or api docs.

prob found here: https://requests.ryanmccue.info/api/source-class-Requests.html#308


ozh:
I think it would be more useful to document more that there are a gazillion of options, give a few pointers and examples and tell where to look at the source, rather than pinpointing precisely how to timeout or every option available.

PR #528, opened 2021-09-15
GH Pages: add QA workflow

jrfnl:
This commit adds a workflow which tests that the GH Pages site can be generated by Jekyll without errors for pull requests to the `gh-pages` branch.

Notes:
* The workflow has got a safeguard in place to prevent these workflows from running on forks which can't deploy the GH Pages website.

### DIFF
--- a/.github/workflows/test-ghpages.yml
+++ b/.github/workflows/test-ghpages.yml
@@ -1,0 +1,27 @@
+name: Test Jekyll Build
+
+on:
+  pull_request:
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  #### TEST THAT THE GH PAGES SITE GETS BUILT WITHOUT ERRORS ####
+  test:
+    name: "Test GHPages site"
+    if: github.repository == 'WordPress/Requests'
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          # Use the version as per https://pages.github.com/versions/.
+          ruby-version: 2.7.3
+          bundler-cache: true
+
+      - name: Test building the GH Pages site
+        run: bundle exec jekyll build


jrfnl:
Note: in the latest version of this PR, I have turned `RECURSIVE` off in the broken link check workflow, but with that turned off, the site doesn't get properly scanned.

However, with it turned on, the scan looks to go into an infinite loop: https://github.com/WordPress/Requests/runs/3605224103?check_suite_focus=true (currently running for 3 hours already)

I also get the impression that the action checks links time and again, even when checked before in the same run.
I've opened an issue about that upstream: https://github.com/technote-space/broken-link-checker-action/issues/131

We may want to remove the broken link workflow from this PR for now until a more stable solution is found.

.github/workflows/test-ghpages.yml
```diff
@@ -0,0 +1,27 @@
+name: Test Jekyll Build
+
+on:
+  pull_request:
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  #### TEST THAT THE GH PAGES SITE GETS BUILD WITHOUT ERRORS ####
```
schlessera:
```suggestion
  #### TEST THAT THE GH PAGES SITE GETS BUILT WITHOUT ERRORS ####
```

.github/workflows/test-ghpages.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          # Use the version as per https://pages.github.com/versions/
```
schlessera:
```suggestion
          # Use the version as per https://pages.github.com/versions/.
```

schlessera (changes requested):

schlessera (approved):

jrfnl:
Note: removed the broken link check for now, we'll revisit later.

Issue #513, opened 2021-08-20
Check through code base for use of "magic numbers"


jrfnl:
Most notably in the cURL transport class, hard-coded numbers are used for various cURL error codes.

Task:
- [x] Go through the code base to find these numbers
- [x] Check if there is a PHP constant which can be used instead and if that constant is available in all supported PHP versions and if so, switch over
- [x] For "magic numbers" which don't have a PHP native constant, create a descriptive class constant instead and use that.

jrfnl:
Relevant links:
* https://curl.se/libcurl/c/libcurl-errors.html
* https://www.php.net/manual/en/curl.constants.php

jrfnl:
PR ready. Will pull after #519 has been merged to prevent merge-conflict hell.

jrfnl:
All related PRs have been merged. Closing the issue.

PR #541, opened 2021-09-17
Requests: set default transports as a constant

jrfnl:
The default supported transports do not change during the request, so should be a constant and as the minimum PHP version is now PHP 5.6, we can use constant arrays.

I've elected to set this array with the same value for key and value to allow using `isset()` against the array.
I've also adjusted the `Requests::add_transport()` method to set the key as well, which automatically allows us to get rid of the expensive and unnecessary `array_merge()`.

The list of supported constants was also used in multiple (duplicate) data providers in the tests.
This duplication has been removed by adding this data provider to the base `TestCase`, making it available to all tests.

The data provider in the base `TestCase`:
* Makes use of the new constant to always provide the list of all supported transport classes.
* Has been set up to provide named data sets to allow for more descriptive tests when running the tests using `--testdox` and for easier debugging.

Part of a PR series to address #513

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -2,8 +2,24 @@
 
 namespace WpOrg\Requests\Tests;
 
+use WpOrg\Requests\Requests;
 use Yoast\PHPUnitPolyfills\TestCases\TestCase as Polyfill_TestCase;
 
 abstract class TestCase extends Polyfill_TestCase {
 
+	/**
+	 * Data provider for use in tests which need to be run against all default supported transports.
+	 *
+	 * @var array
+	 */
+	public function transportProvider() {
+		$data = array();
+
+		foreach (Requests::DEFAULT_TRANSPORTS as $transport) {
+			$name        = substr($transport, (strrpos($transport, '\\') + 1));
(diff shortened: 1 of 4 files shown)


tests/TestCase.php
```diff
@@ -2,8 +2,28 @@
...
+		if (!isset($data)) {
+			$data = array();
+
+			foreach (Requests::DEFAULT_TRANSPORTS as $transport) {
+				$name        = substr($transport, (strrpos($transport, '\\') + 1));
+				$data[$name] = array($transport);
+			}
+		}
+
+		return $data;
```
schlessera: Given that tests are not run in production and thus performance is not that big of a concern, I'd rather keep static state out of tests whenever possible, to avoid potentially having flaky tests because of it.

```suggestion
		$data = array();

		foreach (Requests::DEFAULT_TRANSPORTS as $transport) {
			$name        = substr($transport, (strrpos($transport, '\\') + 1));
			$data[$name] = array($transport);
		}

		return $data;
```

schlessera (changes requested):

schlessera (approved):

PR #542, opened 2021-09-17
InvalidArgument: add new `create()` method

jrfnl:
### InvalidArgument: add new create() method

... to allow for standardizing the messages thrown via this Exception.

Note: the `parent::__construct()` method is not overloaded with a `private` version, so variation is still possible, just not recommended.

Includes adding a perfunctory unit test for the new method.

### DIFF
--- a/src/Exception/InvalidArgument.php
+++ b/src/Exception/InvalidArgument.php
@@ -10,4 +10,32 @@ use InvalidArgumentException;
  * @package Requests
  * @since   2.0.0
  */
-final class InvalidArgument extends InvalidArgumentException {}
+final class InvalidArgument extends InvalidArgumentException {
+
+	/**
+	 * Create a new invalid argument exception with a standardized text.
+	 *
+	 * @param int    $position The argument position in the function signature. 1-based.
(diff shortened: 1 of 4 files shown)


src/Exception/InvalidArgument.php
```diff
@@ -10,4 +10,32 @@
...
+	 * @param int    $position The argument position in the function signature. 1-based.
+	 * @param string $name     The argument name in the function signature.
+	 * @param string $expected The argument type expected as a string.
+	 * @param string $received The actual argument type received.
+	 *
+	 * @return \WpOrg\Requests\Exception\InvalidArgument
+	 */
+	public static function create($position, $name, $expected, $received) {
+		// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace
+		$stack = debug_backtrace();
```
schlessera: To avoid wasting resources, the call should be done like this:
```suggestion
		$stack = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
```

schlessera (changes requested):

schlessera (approved):

PR #545, opened 2021-09-17
GH Actions: add GH Pages update workflow

### LINKED ISSUE #466: GH Pages/GH Actions: create workflow to auto-update the website on a new release
It would be very nice if we could automate regenerating the site whenever a release tag is created on the `master` branch.

Based on the research I have done so far into how the site is set up, that would roughly involve the following steps/conditions:
* Run when a tag has been created on the `master` branch (`release` - `published` event I believe) and the CI for that branch has passed.
    Probably a good idea to also allow manual triggering of the workflow to allow rebuilding of the site intermittently.
* Install PHP at 7.1 or higher.
* Install [ApiGen](https://packagist.org/packages/apigen/apigen) - I suspect at version 4.x for now.
* Update line 15 of the `apigen.neon` file in the `gh-pages` branch to reflect the name of the just released tag.
    The name of the tag should be available in `GITHUB_REF` or alternatively can be obtained via a call to the GH API or via a call to git.
    Refs:
    * https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release
    * https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
    * https://gist.github.com/rponte/fdc0724dd984088606b0
* Run ApiGen from the root of the `gh-pages` branch with the `apigen.neon` config.
    This will update the files in the `gh-pages` `api` directory.
* Install Python
* Run the `buildtool.py` script from the root of the `gh-pages` branch.
    This will update the `index.md` file and the files in the `docs` directory of the `gh-pages` branch.
* Install Ruby 2.7.x
* Run `bundler install` from the `gh-pages` root.
* Test whether the site will build without errors.
* Commit the resulting updated files.
    While working on and testing this workflow, commits should probably go to a `feature` branch so it can be reviewed that the workflow works as expected.
    Once the workflow has stabilized, it should be decided whether the changes should still be committed to a feature branch with the workflow opening a PR so the results will always be reviewed before the site is updated _or_ to allow the workflow to commit to the `gh-pages` branch directly.

I get the impression that the scripting as it is currently assumes a setup with the repo checked out twice like this:
```
project root at `master` branch
  |- bin
  |- docs
  |- examples
  |- library
  |- tests
  |- project root at `gh-pages` branch
    |- __includes
    |- _layouts
    |- .. etc
```

Also note that for some of the above steps/actions, there may well be predefined actions available or even additional actions which would be useful to run.
See: https://github.com/marketplace?type=actions

One action which look interesting in this context:
* https://github.com/marketplace/actions/apigen-docs

For additional actions (future scope), I'm thinking along the lines of:
* Running a spellchecker over the generated site.
* Running a linter which checks for dead links over the generated site.

@rmccue Have you got any additional context which whomever will work on this should know ?

jrfnl:
### GH Pages: switch from APIGen to phpDocumentor

1. APIGen 4.x does not support "modern" PHP, and that includes such things as the PHP 5.5 `::class` syntax.
2. There is an APIGen 5.x RC available, but that version is in its current state is no longer installable.
3. APIGen has not seen a release in over four years, nor a commit in more than three years and is currently unmaintained.

With that in mind, I've taken the decision to switch to the actively maintained phpDocumentor library to generate the API documentation.

This commit actions that decision by:
* Adding a base `phpdoc.dist.xml` configuration file.
* Adding a separate `composer.json` file for use when generating the API docs.
    The `require-dev` will pull in the phpDocumentor Phar via the shim package rather than install the individual dependencies.
* Adding this file and the `build` directory to the `.gitattributes` file for ignoring in distribution zips
* Adding the structure cache directory which will be used and the config overload file to the `.gitignore` file.

In a sister-PR to the `gh-pages` branch, the old `apigen.neon` config file will be removed.

Refs:
* https://github.com/ApiGen/ApiGen/releases
* https://tomasvotruba.com/blog/2017/09/04/how-apigen-survived-its-own-death/
* https://www.phpdoc.org/
* https://github.com/phpDocumentor/phpDocumentor/releases
* https://github.com/phpDocumentor/shim

### GH Pages: add script to update/generate phpDoc config

This commit adds a script which can be run both locally as well as in a GH Actions workflow.

The script:
* Will retrieve the version number of the latest Requests release from the GH API.
* If no `phpdoc.xml` file exists, it will create a `phpdoc.xml` file based on the existing `phpdoc.dist.xml` file and make sure the title contains the correct version number.
* If a `phpdoc.xml` already exists, it will just and only update the title for the docs with the correct version number.

Includes a minor tweak to the PHP `lint` workflow to exclude the scripts in the `build/ghpages` directory from linting on PHP < 7.2.

Open question: once Requests 2.0.0 has been released, should we update the website to only contain the API docs for Requests 2.0.0 or should we keep a copy of the last Requests 1.x docs available as well ?

### GH Pages: add script to update prose docs

This commit adds a script which can be run both locally as well as in a GH Actions workflow.

The script:
* Will take the `README.md` file from the main repo as well as the `md` files in the `/docs/` directory and prepare them for use in the GH Pages website.
    This preparation includes:
    - Adding frontmatter to the files for use by Jekyll.
    - Replacing `md` in links with `html`.
    - Various other tweaks depending on the file.
* The script will, by default, place the adjusted files in the `build/ghpages/artifacts` directory. For local runs where there is already a separate clone of this repo available which has the `gh-pages` branch checked out, a `--target=path/to/ghpages` CLI argument can be passed to place the files directory in the `gh-pages` clone.

Includes a minor tweaks to the `docs/README.md` file for easier processing of that file.

This script effectively replaces the `build.py` script which is in the `gh-pages` branch. That script will be removed in a sister-PR to that branch.

### GH Pages: add a README

... to the new directory containing the scripts documenting how these scripts can be run locally to test what changes would be made when the website would be updated.

### GH Actions: add an "update website" workflow

This new workflow will run when:
* A new release of Requests has been published.
* A pull request has been send in which updates the scripts which are involved in updating the website.

The workflow can also be manually triggered, but will not run in forks of this repo (as they can't update or publish the website).

👉 Open question: should the workflow always run for pushes to `stable` ?
Those will normally be the release pushes and the "next" version nr may not be known yet in that case, but if something is "off" with the website, an update to regen the website would now require a new tag (or manual trigger, but we'll have to see how well that works for this).

The script will:
* In the `prepare` job:
    - (Re-)Generate the API docs using phpDocumentor.
    - Prepare the `README` and the files in the `docs` directory for use in the website.
    - Upload the results as an artifact.
        The artifact will only be available for one day as it's only used in the next job.
* In the `createpr` job:
    - Download the artifact.
    - Remove outdated files and replace them with the newly generated ones.
        Note: files can be updated, as well as added/removed, so for the commits we need to make sure that both tracked as well as untracked changes are committed.
    - Create a PR to the `gh-pages` branch containing two commits:
        1. A commit for the API docs update.
            This commit _may_ be empty if no inline docs have been updated, though I suspect that will be exceedingly rare.
        2. A commit containing the remaining changes (docs + homepage).

For PRs created because the workflow or the associated scripts have changed, the PR title will be prefixed with `[TEST | DO NOT MERGE]` so the effect of workflow/script changes can still be verified, without updating the website with the docs for an unreleased `develop` version of Requests.
For those PRs, the website change PR will be based on the feature branch against `develop` as otherwise the action doesn't have access to the latest version of the workflow/scripts. For releases and manual triggering, the website change PR will always be based on the `stable` branch.

Note: these steps could be combined into one job, but for debugging, it felt cleaner to separate it into two jobs.

👉 Note: we'll need to monitor if a PR correctly gets created when there are no changes to the prose docs (committed via the `create-pr` action), but there are API doc changes (committed manually), I hope it will be, but we'll have to see.

Refs:
* https://github.com/marketplace/actions/create-pull-request
* https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md

Fixes #466

### DIFF
--- a/.github/workflows/update-website.yml
+++ b/.github/workflows/update-website.yml
@@ -1,0 +1,166 @@
+name: Build website
+
+on:
+  # Trigger the workflow whenever a new release is created.
+  release:
+    types:
+      - published
+  # And whenever this workflow or one of the associated scripts is updated.
+  pull_request:
+    paths:
+      - '.github/workflows/update-website.yml'
+      - 'build/ghpages/**'
+  # Also allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  prepare:
+    name: "Prepare website update"
+    # Don't run on forks.
+    if: github.repository == 'WordPress/Requests'
+
+    runs-on: ubuntu-latest
+    steps:
+      # By default use the `stable` branch as the published docs should always
+      # reflect the latest release.
+      # For testing changes to the workflow or the scripts, use the PR branch
+      # to have access to the latest version of the workflow/scripts.
+      - name: Determine branch to use
+        id: base_branch
+        run: |
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            echo '::set-output name=BRANCH::${{ github.ref }}'
+          else
+            echo '::set-output name=BRANCH::stable'
+          fi
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ steps.base_branch.outputs.BRANCH }}
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          ini-values: display_errors=On
+          coverage: none
+
+      # This will install the phpDocumentor PHAR, not the "normal" dependencies.
+      - name: Install Composer dependencies
+        uses: ramsey/composer-install@v1
+        with:
+          composer-options: "--working-dir=build/ghpages/"
+
+      - name: Update the phpDoc configuration
+        run: php build/ghpages/update-docgen-config.php
+
+      - name: Generate the phpDoc documentation
+        run: php build/ghpages/vendor/bin/phpdoc
+
+      - name: Transform the markdown docs for use in GH Pages
+        run: php build/ghpages/update-markdown-docs.php
+
+      # Retention is normally 90 days, but this artifact is only for review
+      # and use in the next step, so no need to keep it for more than a day.
+      - name: Upload the artifacts folder
+        uses: actions/upload-artifact@v2
+        if: ${{ success() }}
+        with:
+          name: website-updates
+          path: ./build/ghpages/artifacts
+          if-no-files-found: error
+          retention-days: 1
+
+  createpr:
+    name: "Create website update PR"
+    needs: prepare
+    # Don't run on forks.
+    if: github.repository == 'WordPress/Requests'
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+
+      - name: Download the prepared artifacts
+        uses: actions/download-artifact@v2
+        with:
+          name: website-updates
+          path: artifacts
+
+      # Different version of phpDocumentor may add/remove files for CSS/JS etc.
+      # Similarly, a new Requests major may remove classes.
+      # So we always need to make sure that the old version of the API docs are
+      # cleared out completely.
+      - name: Clear out the API directory
+        run: rm -vrf ./api/*
+
+      - name: Move the updated API doc files
+        run: mv -fv artifacts/api/* ./api/
+
+      # The commit should contain all changes in the API directory, both tracked and untracked!
+      - name: Commit the API docs separately
+        run: |
+          git config user.name 'GitHub Action'
+          git config user.email '${{ github.actor }}@users.noreply.github.com'
+          git add -A ./api/
+          git commit --allow-empty --message="GH Pages: update API docs for Requests ${{ github.ref }}"
+
+      # Similar to the API docs, files could be removed from the prose docs, so
+      # make sure that the directory is cleared out completely beforehand.
+      - name: Clear out the docs directory
+        run: rm -vf ./docs/*
+
+      - name: Move the other updated files
+        run: |
+          mv -fv artifacts/docs/* ./docs
+          mv -fv artifacts/_includes/* ./_includes
+          mv -fv artifacts/index.md ./index.md
+
+      - name: Verify artifacts directory is now empty
+        run: ls -alR artifacts
+
+      # The directory is also gitignored, but just to be sure.
+      - name: Remove the artifacts directory
+        run: rmdir --ignore-fail-on-non-empty --verbose ./artifacts
+
+      # PRs based on the "pull request" event trigger will contain changes from the
+      # current `develop` branch, so should not be published as the website should
+      # always be based on the latest release.
+      - name: Determine PR title prefix, body and more
+        id: get_pr_info
+        run: |
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            echo '::set-output name=PR_TITLE_PREFIX::[TEST | DO NOT MERGE] '
+            echo '::set-output name=PR_BODY::Test run for the website update after changes to the automated scripts.'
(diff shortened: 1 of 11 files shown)


.github/workflows/update-website.yml
```diff
@@ -0,0 +1,166 @@
...
+      # Similarly, a new Requests major may remove classes.
+      # So we always need to make sure that the old version of the API docs are
+      # cleared out completely.
+      - name: Clear out the API directory
+        run: rm -vrf ./api/*
+
+      - name: Move the updated API doc files
+        run: mv -fv artifacts/api/* ./api/
+
+      # The commit should ocntain all changed in the API directory, both tracked and untracked!
```
schlessera:
```suggestion
      # The commit should contain all changes in the API directory, both tracked and untracked!
```

build/ghpages/README.md
```diff
@@ -0,0 +1,27 @@
+Scripts to update the GH Pages website
+======================================
+
+The scripts in this directory are only for internal use to update the GHPages website associated with this project whenever a new version of the Requests library is released.
```
schlessera: In documentation, we should not assume people know all acronyms.
```suggestion
The scripts in this directory are only for internal use to update the GitHub Pages website associated with this project whenever a new version of the Requests library is released.
```

build/ghpages/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+They are used in the [`update-website.yml`](https://github.com/WordPress/Requests/blob/develop/.github/workflows/update-website.yml) GH Actions workflow.
```
schlessera:
```suggestion
They are used in the [`update-website.yml`](https://github.com/WordPress/Requests/blob/develop/.github/workflows/update-website.yml) GitHub Actions workflow.
```

build/ghpages/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+To run a test build of the GH Pages site locally, execute the following steps:
```
schlessera:
```suggestion
To run a test build of the GitHub Pages site locally, execute the following steps:
```

build/ghpages/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Preparation in this repo:
+* Pre-requisite: use PHP 7.2 or higher.
+* From within this subdirectory, run `composer update -W`.
+* Delete the `build/ghpages/artifacts` subdirectory completely.
+
+Preparation of the GH Pages branch:
```
schlessera:
```suggestion
Preparation of the GitHub Pages branch:
```

build/ghpages/README.md
```diff
@@ -0,0 +1,27 @@
...
+* Delete the `api` directory completely.
+* Delete the `docs` directory completely.
+
+Switch to the project root directory in this clone and:
+* Run `php build/ghpages/update-docgen-config.php` to retrieve the latest tag number from the GH API and create/update the `phpdoc.xml` config.
+* If this was the first time you ran the above script, you now need to edit the `phpdoc.xml` file and update the path in the `<paths>` - `<output>` config to point to the `root/api` directory of the "gh-pages" clone of the repo.
+* Run `php build/ghpages/vendor/bin/phpdoc` to generate the API docs.
+* Run `php build/ghpages/update-markdown-docs.php --target=path/to/gh-pages/root` to generate versions of the markdown docs suitable for use in GH Pages.
+
+You can then use git diff to verify the GH Pages site was updated correctly.
```
schlessera, line 27:
```suggestion
You can then use git diff to verify the GitHub Pages site was updated correctly.
```

schlessera (changes requested):

schlessera (approved):

PR #547, opened 2021-09-17
Introduce new WpOrg\Requests\Utility\InputValidator class

jrfnl:
Includes a complete set of unit tests for the currently added functionality.

### DIFF
--- a/src/Utility/InputValidator.php
+++ b/src/Utility/InputValidator.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * Input validation utilities.
+ *
+ * @package Requests
+ * @subpackage Utilities
+ */
+
+namespace WpOrg\Requests\Utility;
+
+use ArrayAccess;
+use CurlHandle;
+use Traversable;
+
+/**
+ * Input validation utilities.
+ *
+ * @package Requests
+ * @subpackage Utilities
+ */
+final class InputValidator {
+
+	/**
+	 * Verify that a received input parameter is of type string or is "stringable".
+	 *
+	 * @param mixed $input Input parameter to verify.
+	 *
+	 * @return bool
+	 */
+	public static function is_string_or_stringable($input) {
+		return is_string($input) || self::is_stringable_object($input);
+	}
+
+	/**
+	 * Verify whether a received input parameter is usable as an integer array key.
+	 *
+	 * @param mixed $input Input parameter to verify.
+	 *
+	 * @return bool
+	 */
+	public static function is_numeric_array_key($input) {
+		if (is_int($input)) {
+			return true;
+		}
+
+		if (!is_string($input)) {
+			return false;
+		}
+
+		return (bool) preg_match('`^-?[0-9]+$`', $input);
+	}
+
+	/**
+	 * Verify whether a received input parameter is "stringable".
+	 *
+	 * @param mixed $input Input parameter to verify.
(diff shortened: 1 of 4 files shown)


src/Utility/InputValidator.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Verify whether a received input parameter is _accessible as if it were an array_.
+	 *
+	 * @param mixed $input Input parameter to verify.
+	 *
+	 * @return bool
+	 */
+	public static function is_accessible_as_array($input) {
```
schlessera: Potential alternative names:
- `is_indexable()`
- `has_array_access()`

src/Utility/InputValidator.php
```diff
@@ -0,0 +1,111 @@
...
+final class InputValidator {
+
+	/**
+	 * Verify that a received input parameter is of type string or is "stringable".
+	 *
+	 * @param mixed $input Input parameter to verify.
+	 *
+	 * @return bool
+	 */
+	public static function is_string_or_stringable($input) {
```
schlessera, line 30: Potential alternative name:
- `is_stringable()`

src/Utility/InputValidator.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Verify whether a received input parameter is usable as an integer array key.
+	 *
+	 * @param mixed $input Input parameter to verify.
+	 *
+	 * @return bool
+	 */
+	public static function is_numeric_array_key($input) {
```
schlessera, line 41: Potential alternative names:
- `is_decimal()`
- `is_array_index()`

tests/Utility/InputValidatorTest.php
```diff
@@ -0,0 +1,396 @@
...
+					'is_string_or_stringable' => false,
+					'is_numeric_array_key'    => false,
+					'is_stringable_object'    => false,
+					'is_accessible_as_array'  => true,
+					'is_iterable'             => false,
+					'is_curl_handle'          => false,
+				),
+			),
+			'Curl handle (resource or object depending on PHP version)' => array(
+				'input'    => &self::$curl_handle,
```
schlessera: Why is this being passed as an explicit reference here? The way I understand it, it would be a reference for `CurlHandle` anyway, and not matter for a `resource`.

jrfnl: Very good question. Copy/pasta somewhere down the line. Fixed now.

src/Utility/InputValidator.php
jrfnl: That wouldn't be correct as the method strictly checks for a string or Stringable object, while all scalars realistically are _stringable_ (will juggle without errors).

src/Utility/InputValidator.php
jrfnl: Updated in all relevant place to `has_array_access()`.

src/Utility/InputValidator.php
jrfnl: Neither or these feels right...

* `is_decimal()` - floats are also decimal, but not acceptable.
* `is_array_index` - a non-numeric string array index is also a valid array index, but for the purpose of this function should be ignored (return `false`).

schlessera (approved):

Issue #497, opened 2021-06-18
Test: review all tests to stabilize them more


jrfnl:
Currently, there are a lot of tests which follow a pattern along the lines of:

```php
// Do something
$this->assertSame($expected, $result['key']['subkey']->property);
```

Tests like these are unstable as they may error out on any of the following:
* `$result` not being an array.
* The `'key'` index not existing in `$result`
* The `'subkey'` index not existing in `$result['key']`
* ` $result['key']['subkey']` not being an object
* ` $result['key']['subkey']` not being an object of the expected type (instance of)
* The object in `$result['key']['subkey']` not having a property named `$property`.

Those tests should be stabilized and improved by changing them to something along the lines of:

```php
// Do something
$this->assertIsArray($result);
$this->assertHaskey('key', $result);
$this->assertIsArray($result['key']);
$this->assertHaskey('subkey', $result['key']);
$this->assertIsObject($result['key']['subkey']);
$this->assertInstanceOf($classname, $result['key']['subkey']);
$this->assertHasAttribute('property', $result['key']['subkey']);
$this->assertSame($expected, $result['key']['subkey']->property);
```

Additionally, all tests which have multiple assertions, should pass the `$message` parameter to each assertion to help figure out which assertion has failed when a test fails.

### Other review tasks (optional, but highly recommended)

* Refactoring tests methods which do the same thing to data providers.
* Enhancing existing data providers to use named data sets and keyed data entries.
* Adding appropriate `@covers` tags.
* Reviewing code coverage and adding additional tests to raise (both path as well as branch) code coverage.
* Adding test documentation.
* Potentially splitting test classes which do too much into multiple classes based on what is being covered by the tests.

### Actions list

- [x] `Requests\Tests\ChunkedEncodingTest` - PR #581
- [ ] `Requests\Tests\CookiesTest` - PR series, PR #737
    **Owner**: @jrfnl
    **Status**: Ready, waiting to be pulled
- [x] `Requests\Tests\EncodingTest` - PR #595
- [x] `Requests\Tests\IdnaEncoderTest` - PR #549
- [ ] `Requests\Tests\IriTest`
- [ ] `Requests\Tests\RequestsTest`
- [ ] `Requests\Tests\SessionTest`
    **Owner**: @jrfnl
    **Status**: WIP
- [x] `Requests\Tests\SslTest` - PR #551
- [x] `Requests\Tests\Auth\BasicTest` - PR #557
[...]

PR #549, opened 2021-09-17
IdnaEncoderTest: improve tests

jrfnl:
### IdnaEncoderTest: refactor encoding tests

* Merge five tests doing the exact same thing, into one test using a data provider with named data sets and keyed data entries.
* Remove `static` keyword from data provider.
* Add docblock documentation to the test and the data provider.

### IdnaEncoderTest: add additional valid encoding tests

This commit adds:
* An empty text string test case.
* A stringable object test case.
* All other examples listed in the RFC as test cases.

### IdnaEncoderTest: refactor invalid encoding tests

* Merge five tests doing the exact same thing, into one test using a data provider with named data sets.
* Add docblock documentation to the test and the data provider.

### IdnaEncoderTest: minor spacing tweaks

.... to improve readability.

### IdnaEncoder::to_ascii(): add select @codeCoverageIgnore annotation 

Related to #497

### DIFF
--- a/src/IdnaEncoder.php
+++ b/src/IdnaEncoder.php
@@ -92,11 +92,17 @@ class IdnaEncoder {
 		// Step 4: Check if it's ASCII now
 		if (self::is_ascii($text)) {
 			// Skip to step 7
+			/*
+			 * As the `nameprep()` method returns the original string, this code will never be reached until
+			 * that method is properly implemented.
+			 */
+			// @codeCoverageIgnoreStart
 			if (strlen($text) < self::MAX_LENGTH) {
 				return $text;
 			}
 
 			throw new Exception('Prepared string is too long', 'idna.prepared_too_long', $text);
+			// @codeCoverageIgnoreEnd
 		}
 
 		// Step 5: Check ACE prefix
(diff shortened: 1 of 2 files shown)


jrfnl:
Reverted to draft as this PR needs PR #547 to be merged first.

src/IdnaEncoder.php
```diff
@@ -92,11 +92,17 @@ public static function to_ascii($string) {
 		// Step 4: Check if it's ASCII now
 		if (self::is_ascii($string)) {
 			// Skip to step 7
+			/*
+			 * As the `nameprep()` method returns the original string, this code will never be reached until
+			 * that method is properly implemented.
+			 */
+			// @codeCoverageIgnoreStart
```
schlessera, line 99: I think this would be clearer if the entire construct is a single comment entity.
```suggestion
			/*
			 * As the `nameprep()` method returns the original string, this code will never be reached until
			 * that method is properly implemented.
			 *
			 * @codeCoverageIgnoreStart
			 */
```

schlessera (changes requested):

src/IdnaEncoder.php
jrfnl: @schlessera I 100% agree. Unfortunately, that doesn't work.

image

versus

image

jrfnl:
Marking this as ready for review now PR #547 has been merged.

schlessera (approved):

PR #563, opened 2021-09-19
GH Actions/update website: minor tweak

jrfnl:
The test PR created a new label `"Type: documentation"` due to the quotes used in the script. This removes the quotes, though it remains to be seen if that works correctly as it _may_ confuse Yaml due to the `:` colon in the label.

If this doesn't work, the issue should probably be reported upstream to the `peter-evans/create-pull-request` project.


jrfnl:
Based on #564, this fix works.

schlessera (approved):

PR #572, opened 2021-10-11
Ssl: add input validation to all methods

jrfnl:
None of the methods in the `WpOrg\Requests\Ssl` class did any input validation. This could lead to various PHP errors and becomes all the more relevant due to the new "passing null to non-nullable" deprecation notice on PHP 8.1.

This commit adds simple input validation to all methods within the `WpOrg\Requests\Ssl` class and will throw an `WpOrg\Requests\Exception\InvalidArgument` exception if a parameter does not comply with the expectations.

Note: for the `$cert` array parameter in the `Ssl::verify_certificate()` method, I considered adding an `array` type declaration, however, for consistency, always throwing the Requests native `InvalidArgument` exception seemed the better choice.

Includes adding tests for the new exception.

### DIFF
--- a/src/Ssl.php
+++ b/src/Ssl.php
@@ -8,6 +8,9 @@
 
 namespace WpOrg\Requests;
 
+use WpOrg\Requests\Exception\InvalidArgument;
+use WpOrg\Requests\Utility\InputValidator;
+
 /**
  * SSL utilities for Requests
  *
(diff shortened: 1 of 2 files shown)


src/Ssl.php
```diff
@@ -118,8 +136,13 @@ public static function verify_reference_name($reference) {
 	 * @param string $host Requested host
 	 * @param string $reference dNSName to match against
 	 * @return boolean Does the domain match?
+	 * @throws \WpOrg\Requests\Exception\InvalidArgument When the either of the passed arguments is not a string or a stringable object.
```
schlessera:
```suggestion
	 * @throws \WpOrg\Requests\Exception\InvalidArgument When either of the passed arguments is not a string or a stringable object.
```

schlessera (changes requested):

src/Ssl.php
jrfnl: Thanks for catching that! Fixed now.

schlessera (approved):

PR #574, opened 2021-10-11
Auth\Basic: add input validation to constructor

jrfnl:
The `Basic::__construct()` method expects to receive an array of exactly two element and did validate the input received, but only threw an exception when the passed parameter did not match the expected element count, not when the passed parameter wasn't an array.

At the same time, the default value of `$args` is `null`, not an empty array and the `$user` and `$pass` properties are both public.

The only reason I can think of to have these properties `public` and the default value set to `null`, is to allow for delayed setting of the `$user` and `$pass`, after the class has already been instantiated.

To maintain the existing behaviour, but still add input validation, an exception is now thrown when the `$args` parameter is neither an array, nor `null`.

Includes;
* A dedicated test for the input validation.
* An additional test with Basic authentication with an instantiated class, but with delayed setting of the `$user` and `$pass` directly on the instantiated object. This test is solely added to document that this is a supported feature (and to safeguard against regressions in that respect).

### DIFF
--- a/src/Auth/Basic.php
+++ b/src/Auth/Basic.php
@@ -10,6 +10,7 @@ namespace WpOrg\Requests\Auth;
 
 use WpOrg\Requests\Auth;
 use WpOrg\Requests\Exception;
+use WpOrg\Requests\Exception\InvalidArgument;
 use WpOrg\Requests\Hooks;
 
 /**
@@ -39,8 +40,10 @@ class Basic implements Auth {
 	/**
 	 * Constructor
 	 *
-	 * @throws \WpOrg\Requests\Exception On incorrect number of arguments (`authbasicbadargs`)
 	 * @param array|null $args Array of user and password. Must have exactly two elements
+	 *
+	 * @throws \WpOrg\Requests\Exception\InvalidArgument When the passed argument is not an array or null.
+	 * @throws \WpOrg\Requests\Exception                 On incorrect number of arguments (`authbasicbadargs`).
 	 */
 	public function __construct($args = null) {
 		if (is_array($args)) {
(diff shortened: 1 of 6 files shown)


src/Auth/Basic.php
```diff
@@ -39,8 +40,10 @@ class Basic implements Auth {
 	/**
 	 * Constructor
 	 *
-	 * @throws \WpOrg\Requests\Exception On incorrect number of arguments (`authbasicbadargs`)
 	 * @param array|null $args Array of user and password. Must have exactly two elements
+	 *
+	 * @throws \WpOrg\Requests\Exception\InvalidArgument When the passed argument is not an array or null.
+	 * @throws \WpOrg\Requests\Exception                 On incorrect number of arguments (`authbasicbadargs`).
```
schlessera: Should we immediately change this Exception into one that is more precise? I think it should be based on https://www.php.net/manual/en/class.badmethodcallexception.php. If we make sure that it extends `\WpOrg\Requests\Exception`, it shouldn't be a breaking change.

schlessera (changes requested):

src/Auth/Basic.php
jrfnl: Good idea. What about making it an `ArgumentCountException` though ? Kind of feels more appropriate ?
https://www.php.net/manual/en/class.argumentcounterror.php

schlessera: That doesn't exist for PHP 5.6, though...

jrfnl: > That doesn't exist for PHP 5.6, though.

True, but that shouldn't really matter if we extend the `\WpOrg\Requests\Exception` class anyhow.

> If we make sure that it extends `\WpOrg\Requests\Exception`, it shouldn't be a breaking change.

jrfnl: I've added the `ArgumentCount` exception class + implemented use of it in this class + one other location in the code base where it seemed appropriate.

schlessera (approved):

PR #571, opened 2021-10-09
Remove expired DST Root CA X3 from bundled cacert.pem

wojsmol:
related https://github.com/WordPress/Requests/issues/568

## Pull Request Type

- [x] I have checked there is no other PR open for the same change.

This is a:
- [x] Bug fix
- [ ] New feature
- [ ] Code quality improvement

## Context
[WordPress core commit in develop](https://github.com/WordPress/wordpress-develop/commit/3ae66a3cd1b262d40930b47432c10a4f3babe1c3)
[Old Let’s Encrypt Root Certificate Expiration and OpenSSL 1.0.2](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/)

## Detailed Description
This PR  restores Requests compatibility with servers running openssl 1.0.2 by default see openssl project blog linked above. 
## Quality assurance

- [x] This change does NOT contain a breaking change (fix or feature that would cause existing functionality to change).
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added unit tests to accompany this PR.
- [x] The (new/existing) tests cover this PR 100%.
- [x] I have (manually) tested this code to the best of my abilities.
- [x] My code follows the style guidelines of this project.

## Documentation
Document DST Root CA X3 root certificate removal in changelog

PR #577, opened 2021-10-12
update cacert to the latest version

ZsgsDesign:
## Pull Request Type

This is a:
- [x] Bug fix
- [ ] New feature

## Context

This PR updates `cacert.pem` following #571.

When I failed with some requests today, I checked the code, finding that the reason is Let's Encrypt. Their old root CA expired last month and their new root CA ISRG Root X1 still contains the information this it was issued by the now-expired DST Root CA X3, this trusted path is checked (until the end of the chain) and then fails.

So it is of utter importance to remove it now and update a new patch as `1.8.2`.

Also because this project uses its own cacert, we should update the cacert regularly.

This PR update cacert to a most recent version of it, released on 30 Sep, as #571 did not alter the date and hash and that is not recommended.

## Quality assurance


schlessera:
Thanks for the PR, @ZsgsDesign !

We've opted for a fresh pull with no manual changes, and we'll be looking at automating this updating in the future via GHA.

PR #589, opened 2021-11-05
Requests: change `$magic_compression_headers` property to constant

jrfnl:
Follow up on #309

As this is a non-writable, constant array which should never be changed during runtime, this array can be changed from a property to a class constant, as supported since PHP 5.6.

### DIFF
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -140,6 +140,25 @@ class Requests {
 	);
 
 	/**
+	 * All (known) valid deflate, gzip header magic markers.
+	 *
+	 * These markers relate to different compression levels.
+	 *
+	 * @link https://stackoverflow.com/a/43170354/482864 Marker source.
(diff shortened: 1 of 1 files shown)


jrfnl:
Ha! Now I remember why I didn't implement it like this originally 😁 

```php
isset(self::MAGIC_COMPRESSION_HEADERS[$marker])
```
... is not allowed until PHP 7.0 and generates a `Fatal error: Cannot use isset() on the result of an expression ... in ./src/Requests.php on line 846` in PHP 5.6.....

This should be addressed once the minimum PHP version has gone up to PHP 7.0 (or higher).

jrfnl:
Note: the same applies to the `WpOrg\Requests\Autoload::$deprecated_classes` array, which can't be changed to a class constant for the same reason.

PR #592, opened 2021-11-06
IdnaEncoder::encode(): add input validation

jrfnl:
The `IdnaEncoder::encode()` is the official entry point for this class, even though there are more public methods (of which it is questionable whether they should be public, but changing that now would be a BC-break).

The documented accepted parameter type is `string`, but no input validation was done on the parameter, which could lead to various PHP errors, most notably a "passing null to non-nullable" deprecation notice on PHP 8.1.

This commit adds input validation to the `IdnaEncoder::encode()` method, allowing only for strings and _stringable_ objects.

While this is stricter than previously, only a string could result in a valid return value, so in my opinion, being strict is warranted.

Includes adding perfunctory tests for the new exception.

### DIFF
--- a/src/IdnaEncoder.php
+++ b/src/IdnaEncoder.php
@@ -3,6 +3,8 @@
 namespace WpOrg\Requests;
 
 use WpOrg\Requests\Exception;
+use WpOrg\Requests\Exception\InvalidArgument;
+use WpOrg\Requests\Utility\InputValidator;
 
 /**
  * IDNA URL encoder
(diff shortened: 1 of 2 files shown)


jrfnl:
**Open question**: Should this method/class allow for `Stringable` objects ? The current implementation allows for it, but we may want to elect not to support `Stringable` objects as input.

schlessera:
> Should this method/class allow for `Stringable` objects ?

I'd vote for yes, as there is no technical reason not to accept them.

schlessera (approved):

jrfnl:
> > Should this method/class allow for `Stringable` objects ?
> 
> I'd vote for yes, as there is no technical reason not to accept them.

Well, the technical reason not to accept them would be that we won't be able to add type declarations to functions accepting `string|Stringable` until PHP 8.0 (i.e. more likely to happen in Requests 4.0 or 5.0, not 3.0)

Issue #533, opened 2021-09-17
PHP 8.0: full review for named parameters


jrfnl:
PHP 8.0 introduced named parameters and the code in Requests needs to be reviewed to be compatible with this.

Ref:
* https://wiki.php.net/rfc/named_params

The following two reviews should be done:
- [x] Ensure that (child) classes use the same parameter names as their PHP native parent classes/interfaces
- [x] Ensure that (child) classes use the same parameter names as their Requests native parent classes/interfaces
- [ ] Review the remaining parameter names for being sensible.
- [x] Include ensuring that reserved keywords are not being used as parameter names to prevent confusion.

jrfnl:
Closing as most tasks have been done in detail.

The "generic" parameter review has been done in a more cursory manner during the addressing of the input validation (#593).

Any further parameter renames are now bound to major releases.

PR #599, opened 2021-11-07
PHP 8.0 | Hooks::dispatch(): prevent potential fatal error

jrfnl:
### PHP 8.0 | Hooks::dispatch(): prevent potential fatal error

This commit adds a safeguard on the off-chance that the `Hooks::dispatch()` method would ever be called with an associative array for the parameters.

This prevents a potential fatal `Unknown named parameter ...` error in PHP 8.x.

Includes unit test.

Related to #533

### 🆕  Hooks::dispatch(): modernize dynamic function call

... by replacing the `call_user_func_array()` with a variable function call in combination with the spread operator.

Includes adding a dedicated test to safeguard that this change does not break the handling of references when passed as part of the `$parameters` to dispatch.

### DIFF
--- a/src/Hooks.php
+++ b/src/Hooks.php
@@ -88,7 +88,7 @@ class Hooks implements HookManager {
 
 		foreach ($this->hooks[$hook] as $priority => $hooked) {
 			foreach ($hooked as $callback) {
-				call_user_func_array($callback, $parameters);
+				$callback(...$parameters);
 			}
 		}
 
--- a/src/Hooks.php
+++ b/src/Hooks.php
@@ -81,6 +81,11 @@ class Hooks implements HookManager {
 			return false;
 		}
 
+		if (!empty($parameters)) {
+			// Strip potential keys from the array to prevent them being interpreted as parameter names in PHP 8.0.
+			$parameters = array_values($parameters);
+		}
+
 		foreach ($this->hooks[$hook] as $priority => $hooked) {
 			foreach ($hooked as $callback) {
 				call_user_func_array($callback, $parameters);
--- a/tests/HooksTest.php
+++ b/tests/HooksTest.php
@@ -203,6 +203,32 @@ class HooksTest extends TestCase {
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


src/Hooks.php
```diff
@@ -83,6 +83,11 @@ public function dispatch($hook, $parameters = array()) {
...
 		}
 
+		if (!empty($parameters)) {
+			// Strip potential keys from the array to prevent them being interpreted as parameter names in PHP 8.0.
+			$parameters = array_values($parameters);
+		}
+
 		foreach ($this->hooks[$hook] as $priority => $hooked) {
 			foreach ($hooked as $callback) {
 				call_user_func_array($callback, $parameters);
```
schlessera: Why are we not using variadic arguments here instead?
```suggestion
				$callback(...$parameters);
```

jrfnl: > Why are we not using variadic arguments here instead?

Historically because it wasn't possible previously as the minimum PHP version did not allow for the spread operator ;-)

As for changing it now... I'm a bit wary of doing so as the behaviour will change with regards to references and the method is called quite often with variables being passed by reference.

Also see https://core.trac.wordpress.org/ticket/47678#comment:31 - section 3.

jrfnl: Then again... `array_values()` may already break the references...

schlessera: Yes, that's what I thought too. If we are using `array_values()`, it will be pretty much the same thing as using the variadic arguments operator.

schlessera: Well, I'm not 100% sure, I think this needs to be tested.

jrfnl: I've tested this and added a unit test to safeguard this as well.

The test passed prior to the code changes and still passes with both the code changes in place, so I think we're good.

jrfnl: (to my surprise, I may add)

jrfnl:
FYI: re-pushed commit has no functional changes, just added the `$message` parameter to the assertions as there are multiple assertions in the `testDispatchHandlesParametersPassedByReference()` test.

schlessera (approved):

PR #608, opened 2021-11-12
Fix reports by PHPStan

szepeviktor:
Fix things while running `vendor/bin/phpstan analyze src/ -l 3`

## Pull Request Type

- [ ] I have checked there is no other PR open for the same change.

This is a:
- [x] Bug fix
- [ ] New feature

### DIFF
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -621,7 +621,7 @@ class Requests {
 	 * @param array|null $data Data to send either as a query string for GET/HEAD requests, or in the body for POST requests
 	 * @param string $type HTTP request type
 	 * @param array $options Options for the request
-	 * @return void $options is update with the results
+	 * @return void $options is updated with the results
 	 */
 	protected static function set_defaults(&$url, &$headers, &$data, &$type, &$options) {
 		if (!preg_match('/^http(s)?:\/\//i', $url, $matches)) {
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -621,7 +621,7 @@ class Requests {
 	 * @param array|null $data Data to send either as a query string for GET/HEAD requests, or in the body for POST requests
 	 * @param string $type HTTP request type
 	 * @param array $options Options for the request
-	 * @return void
+	 * @return void $options is update with the results
 	 */
 	protected static function set_defaults(&$url, &$headers, &$data, &$type, &$options) {
 		if (!preg_match('/^http(s)?:\/\//i', $url, $matches)) {
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -621,7 +621,7 @@ class Requests {
 	 * @param array|null $data Data to send either as a query string for GET/HEAD requests, or in the body for POST requests
 	 * @param string $type HTTP request type
 	 * @param array $options Options for the request
-	 * @return array $options
+	 * @return void
 	 */
 	protected static function set_defaults(&$url, &$headers, &$data, &$type, &$options) {
 		if (!preg_match('/^http(s)?:\/\//i', $url, $matches)) {
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -793,7 +793,7 @@ class Requests {
 	 *
 	 * @param string $response Full response text including headers and body (will be overwritten with Response instance)
 	 * @param array $request Request data as passed into {@see \WpOrg\Requests\Requests::request_multiple()}
-	 * @return null `$response` is either set to a \WpOrg\Requests\Response instance, or a \WpOrg\Requests\Exception object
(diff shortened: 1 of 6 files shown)


szepeviktor:
Please consider using @phpstan
It will show errors in the code (not in PHPDoc blocks).

src/Requests.php
```diff
@@ -621,7 +621,7 @@ public static function set_certificate_path($path) {
 	 * @param array|null $data Data to send either as a query string for GET/HEAD requests, or in the body for POST requests
 	 * @param string $type HTTP request type
 	 * @param array $options Options for the request
-	 * @return array $options
+	 * @return void
```
jrfnl: While technically true, the `$options` parameter will be updated by reference, so I think if this is changed, that should be made clearer elsewhere in the docblock.

src/Utility/CaseInsensitiveDictionary.php
```diff
@@ -76,7 +76,7 @@ public function offsetGet($offset) {
 	 *
 	 * @throws \WpOrg\Requests\Exception On attempting to use dictionary as list (`invalidset`)
 	 *
-	 * @param string $offset Item name
+	 * @param string|null $offset Item name
```
jrfnl: This is incorrect. Offset is not allowed to be `null` and an exception is thrown when it is.

This has to do with the fact that `ArrayAccess` is implemented and that this class explicitly does not support adding to this class via `$array[] = ...`.

jrfnl (changes requested):
Thanks @szepeviktor. Happy to accept these changes once my remarks are addressed.

Adding PHPStan is something for "maybe later" as the current focus is on getting version 2.0.0 released without further delay.

src/Utility/CaseInsensitiveDictionary.php
szepeviktor: > Offset is not allowed to be `null`

Not **allowed**. You see PHPDoc from the point of view of the user/IDE. I see it as it is written in line number 84.
Please tell me what to do. I'm not able to let go technical facts: thus not able to write contradicting PHPDoc lines.

szepeviktor: All right. I know WordPress users are not seasoned developers. And it is not typehinted as WP supports PHP 5.
It just makes me stop contributing.

src/Requests.php
```diff
@@ -621,7 +621,7 @@ public static function set_certificate_path($path) {
 	 * @param array|null $data Data to send either as a query string for GET/HEAD requests, or in the body for POST requests
 	 * @param string $type HTTP request type
 	 * @param array $options Options for the request
-	 * @return array $options
+	 * @return void $options is update with the results
```
jrfnl:
```suggestion
	 * @return void $options is updated with the results
```

src/Utility/CaseInsensitiveDictionary.php
jrfnl: If I could add the `string` type declaration, I would, but I can't with the current minimum PHP requirements. And even if the minimum PHP version would be upped, I still couldn't, as it would violate the Liskov principle of contravariance for parameters what with the [`ArrayAccess`](https://www.php.net/manual/en/class.arrayaccess.php) interface declaring the parameter as `mixed`.

So, yes, technically `null` is allowed - or rather `mixed` is allowed -, but that goes against the whole point of the class which in reality only supports integer and string keys (as those are the only array keys supported by PHP itself) and is only _intended_ to be used with string keys.

So I'd rather the docs communicate the type we **want** to receive than to get complaints about the Exception being thrown when `null` would be passed.

jrfnl: If I could add the `string` type declaration, I would, but I can't with the current minimum PHP requirements. And even if the minimum PHP version would be upped, I still couldn't, as it would violate the Liskov principle of contravariance for parameters what with the [`ArrayAccess`](https://www.php.net/manual/en/class.arrayaccess.php) interface declaring the parameter as `mixed`.

So, yes, technically `null` is allowed - or rather `mixed` is allowed -, but that goes against the whole point of the class which in reality only supports integer and string keys (as those are the only array keys supported by PHP itself) and is only _intended_ to be used with string keys.

So I'd rather the docs communicate the type we **want** to receive than to get complaints about the Exception being thrown when `null` would be passed.

szepeviktor:
@jrfnl I leave all this up to you. Please close it or merge it.

jrfnl (approved):
Thanks @szepeviktor ! The current changes as they are LGTM.

szepeviktor:
You're welcome.

PR #617, opened 2021-11-15
Requests: remove duplicate `match_domain()` method

jrfnl:
This method was "moved" to the `SSL` class in 2013 via 0c2fe6e24e7ec34117b3bbd3d227046920c92c86 and is publicly accessible in that class. The method within the `Requests` class is never called from within the Requests code base and shouldn't exist.

### DIFF
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -1036,25 +1036,4 @@ class Requests {
 
 		return false;
 	}
-
-	public static function match_domain($host, $reference) {
-		// Check for a direct match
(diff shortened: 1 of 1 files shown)


rashidkalwar786:
this method will work.

jrfnl:
> this method will work.

@rashidkalwar786 Sorry, it is not clear to me what you mean to say with your feedback.

schlessera (approved):

PR #632, opened 2021-11-16
Update bundled certificates

jrfnl:
Updated with the latest version as per Tue Oct 26 03:12:05 2021 GMT.
Source: https://curl.se/docs/caextract.html

This update is a plain include and does not follow the previous philosophy of not always removing expired certificates straight away.

Closes #577


schlessera (approved):

wojsmol:
> This update is a plain include and does not follow the previous philosophy of not always removing expired certificates straight away.

If so then we van automate this with GHA.

Issue #220, opened 2016-06-20
Missing Available Hook in Doc and Missing Hook Parameter


mamist2790:
There is a hook, 'curl.after_send', located in Requests/Transport/cURL.php on line 165 that is not listed in the Hooks section of the documentation. Also, the parameters array is empty and I believe it should include &$this->handle.

For example in my case I need to have access to the request headers in case of an error so I can log them. I set CURLINFO_HEADER_OUT using curl_setopt in my curl.before_send hook. In order to actually access that header though I still need access to the handle when using the curl.after_send hook.

jrfnl:
Before we actually action this, we should have a good look at the security implications of allowing access to the `$handle` (not just here).

Issue #321, opened 2018-03-30
$info parameter isn't passed to <transport>.after_request hook on non-blocking requests


johnbillion:
The `fsockopen.after_request` and `curl.after_request` hooks don't receive an `$info` parameter when a non-blocking request is sent.

This means the hook signature differs depending on whether the request is blocking or not, and conflicts with the [hook docs](https://github.com/rmccue/Requests/blob/master/docs/hooks.md).

Possible solution: Pass an empty array as the `$info` parameter for these hooks when a non-blocking request is performed.

Since #206.

jrfnl:
@johnbillion I've done a full review of the Hooks documentation today, with PR #646 as a result. While this doesn't change the way the hook is being dispatched, it does update the documentation to be more informative regarding this.

If you have the time, would be great if you could look the PR over and give your opinion.

jrfnl:
This proposal is being discussed in #646. Copying the relevant parts of that conversation here for visibility:

@schlessera:
> Hmm, still on the fence about this one, as I'm not yet sure whether the documentation for the optional `$info` args is sufficient here. It might be more robust to go with @johnbillion's suggestion and provide an `$info` array in all cases, with an empty one being used for non-blocking requests. I'll think about this some more...

@jrfnl:
> Deciding whether or not to always pass `$info` is a different issue and if/when it is so decided, the documentation can be further updated to reflect that change.
> 
> Generally speaking I'm in favour of John's suggestion and would recommend for hooks to have the same number of arguments every time they are called, however, making this change now _could_ be considered a breaking change as functions which "hook in" very likely will need updating (callbacks may have a different default value set - `null` - for the optional parameter from the empty array we would pass, callbacks may use `isset()` on the result of `function_get_args()` for this parameter, which would break, etc).

PR #646, opened 2021-11-25
Docs: full review of Hook documentation

jrfnl:
Inspired by issue #220, I've:
* Verified that all hooks listed in the documentation still exist in the codebase.
* Verified the parameters for all hooks listed in the docs with the actual parameters in the hook calls in the codebase.
* Verified that all hooks in the codebase are listed in the Hooks documentation page, with the exception of the `transport.internal....` hooks.

Note:
* The hook descriptions for the added hooks can probably use further improvement. I wasn't always sure how best to describe the hook event.
* The ordering of this list looks to be a loose "requests hooks, curl hooks, fsockopen hooks and those in the order they are called" order.
    With some of the additional hooks, it wasn't always clear where to place them within this loose order, so suggestions to improve the order are welcome.

Also note that the naming of the `request.progress` hook seems inconsistent. I do understand why the hook uses `request` instead of `requests`, but this may be something to revisit at a later point in time.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -16,45 +16,116 @@ Available Hooks
 
     Alter the request before it is sent to the transport.
 
-    Parameters: `string &$url`, `array &$headers`, `array|string &$data`,
+    Parameters: `string &$url`, `array &$headers`, `array|null &$data`,
     `string &$type`, `array &$options`
+
+* **`request.progress`**
+
+    Run events based on how much body data has been received up to this point for the current request.
+
+    Parameters: `string $data`, `int $response_bytes`, `int|bool $response_byte_limit`
 
 * **`requests.before_parse`**
 
     Alter the raw HTTP response before parsing.
 
-    Parameters: `string &$response`
+    Parameters: `string &$response`, `string $url`, `array $headers`,
+    `array|null $data`, `string $type`, `array $options`
+
+* **`requests.before_redirect_check`**
+
+    Alter the response object before a potential redirect is detected and acted on.
+
+    Parameters: `WpOrg\Requests\Response &$return`, `array $req_headers`,
+    `array $req_data`, `array $options`
+
+* **`requests.before_redirect`**
+
+    Alter the request arguments before a redirect is acted upon.
+
+    Parameters: `string &$location`, `array &$req_headers`, `array &$req_data`,
+    `array &$options`, `WpOrg\Requests\Response $return`
 
 * **`requests.after_request`**
 
     Alter the response object before it is returned to the user.
 
-    Parameters: `WpOrg\Requests\Response &$return`
+    Parameters: `WpOrg\Requests\Response &$return`, `array &$req_headers`,
+    `array &$req_data`, `array &$options`
+
+* **`multiple.request.complete`**
+
+    Alter the response for an individual request in a multi-request.
+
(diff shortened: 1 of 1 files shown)


jrfnl:
Also related to #321

schlessera (approved):

schlessera:
Hmm, still on the fence about this one, as I'm not yet sure whether the documentation for the optional `$info` args is sufficient here. It might be more robust to go with @johnbillion's suggestion and provide an `$info` array in all cases, with an empty one being used for non-blocking requests. I'll think about this some more...

jrfnl:
> Hmm, still on the fence about this one, as I'm not yet sure whether the documentation for the optional `$info` args is sufficient here. It might be more robust to go with @johnbillion's suggestion and provide an `$info` array in all cases, with an empty one being used for non-blocking requests. I'll think about this some more...

Those are two separate issues to address though. The documentation is issue 1, at least with this PR it would be "hook complete", which is much better then it is now.

Deciding whether or not to always pass `$info` is a different issue and if/when it is so decided, the documentation can be further updated to reflect that change.

Generally speaking I'm in favour of John's suggestion and would recommend for hooks to have the same number of arguments every time they are called, however, making this change now _could_ be considered a breaking change as functions which "hook in" very likely will need updating (callbacks may have a different default value set - `null` - for the optional parameter from the empty array we would pass, callbacks may use `isset()` on the result of `function_get_args()` for this parameter, which would break, etc).

docs/hooks.md
```diff
@@ -16,46 +16,115 @@ Available Hooks
...
 
 * **`curl.after_request`**
 
     Alter the raw HTTP response before returning for parsing.
 
-    Parameters: `string &$response, array &$info`
+    Parameters: `string &$headers`, `[array &$info]`
+
+    The optional `$info` parameter contains the associated array as defined in
+    the return value for [curl_getinfo()](https://www.php.net/curl-getinfo#refsect1-function.curl-getinfo-returnvalues).
```
schlessera, line 93:
```suggestion

This optional parameter will be present when a blocking request was made (`$options['blocking' = true`) and will not be present when a non-blocking request was made (`$options['blocking' = false`). The callback signature needs to be adapted accordingly.

```

docs/hooks.md
```diff
@@ -70,16 +139,18 @@ Available Hooks
...
 
     Alter the raw HTTP response before returning for parsing.
 
-    Parameters: `string &$response, array &$info`
-
-    `$info` contains the associated array as defined in the return value for [stream_get_meta_data()](https://www.php.net/stream-get-meta-data#refsect1-function.stream-get-meta-data-returnvalues).
+    Parameters: `string &$headers`, `[array &$info]`
 
+    The optional `$info` parameter contains the associated array as defined
+    in the return value for [stream_get_meta_data()](https://www.php.net/stream-get-meta-data#refsect1-function.stream-get-meta-data-returnvalues).
```
schlessera, line 156:
```suggestion

This optional parameter will be present when a blocking request was made (`$options['blocking' = true`) and will not be present when a non-blocking request was made (`$options['blocking' = false`). The callback signature needs to be adapted accordingly.

```

schlessera (changes requested):

schlessera (approved):

PR #647, opened 2021-11-25
Hooks: priority should be respected

### LINKED ISSUE #452: Hook priority not working
Sorry in advance for my english

I use hook in order to change cURL options
As the bundle add its own hook, I add a priority nomber in order to run mine later.
Unfortunately, it hasn't changed anything.

After looking in the, I found the issue.
In the script of the hook, the priority is used as key for an array of hook.
However, even if the key is a number, the array is not sorted with it.

As solution, I propose to simply use ksort in the register function :
```php
public function register($hook, $callback, $priority = 0) {
	if (!isset($this->hooks[$hook])) {
		$this->hooks[$hook] = array();
	}
	if (!isset($this->hooks[$hook][$priority])) {
		$this->hooks[$hook][$priority] = array();
		ksort($this->hooks[$hook]);
	}
	$this->hooks[$hook][$priority][] = $callback;
}
```

jrfnl:
## Pull Request Type

This is a:
- [x] Bug fix
- [ ] New feature

## Detailed Description

The `$priority` parameter in the `register()` method is intended to allow for prioritizing the order in which callbacks are run, however, this priority was effectively not used, so callbacks would be run in the order different priorities were _registered_, not ordered _by_ the priorities.

This commit fixes this + adds a unit test to safeguard the fix.

I've elected to add the sorting to the `dispatch()` method, rather than the `register()` method for performance reasons.
Sorting arrays can be expensive and this way, only the subarrays of those hooks which are actually being run will be sorted, instead of sorting the subarray every time a new callback is registered, whether the hook on which the callback is registered will be called or not.

Fixes #452

## Quality assurance

## Documentation

For new features:

### DIFF
--- a/src/Hooks.php
+++ b/src/Hooks.php
@@ -86,6 +86,8 @@ class Hooks implements HookManager {
 			$parameters = array_values($parameters);
 		}
 
+		ksort($this->hooks[$hook]);
+
 		foreach ($this->hooks[$hook] as $priority => $hooked) {
 			foreach ($hooked as $callback) {
 				$callback(...$parameters);
--- a/tests/HooksTest.php
+++ b/tests/HooksTest.php
@@ -242,6 +242,73 @@ class HooksTest extends TestCase {
 	}
 
 	/**
+	 * Verify that hooks are executed based on their priority order.
+	 *
+	 * Issue https://github.com/WordPress/Requests/issues/452
+	 *
+	 * @covers ::dispatch
+	 *
+	 * @return void
+	 */
+	public function testDispatchRespectsHookPriority() {
+		// Register multiple callbacks for the same hook with a variation of priorities.
+		$this->hooks->register(
+			'hook_a',
+			function(&$text) {
+				$text .= "no prio 0\n";
+			}
+		);
+		$this->hooks->register(
+			'hook_a',
+			function(&$text) {
+				$text .= "prio 10-1\n";
+			},
+			10
+		);
+		$this->hooks->register(
+			'hook_a',
+			function(&$text) {
+				$text .= "prio -3\n";
+			},
+			-3
+		);
+		$this->hooks->register(
+			'hook_a',
+			function(&$text) {
+				$text .= "prio 5\n";
+			},
+			5
+		);
+		$this->hooks->register(
+			'hook_a',
+			function(&$text) {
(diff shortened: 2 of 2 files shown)


src/Hooks.php
```diff
@@ -86,6 +86,8 @@ public function dispatch($hook, $parameters = []) {
 			$parameters = array_values($parameters);
 		}
 
+		ksort($this->hooks[$hook]);
```
schlessera, line 89: I don't think this is stable for hooks with the same priority for PHP < 8.0. So, for example for the below tests, hooks `10-1` and `10-2` are not guaranteed to always stay in that same order after `ksort()`, as the behavior for same keys is undefined. From the PHP manual at https://www.php.net/manual/en/function.ksort.php:
> Note:
>
>If two members compare as equal, they retain their original order. Prior to PHP 8.0.0, their relative order in the sorted array was undefined.

Here's how I solved this in a different project: https://github.com/ampproject/amp-toolbox-php/blob/b7377775da193e9cd93fb904421181b47134af70/src/Validator/ValidationErrorCollection.php#L61-L83

I think we'll need something like this here as well, if we want to guarantee that hooks inserted in a certain order for a same priority stay in that same order.

schlessera (changes requested):

src/Hooks.php
jrfnl: The thing is, there never was any mention (documentation) or guarantee that hooks at the _same_ priority would run in a certain order. So yes, while this _may_ change behaviour for callbacks registered on the _same_ priority, the order for those was never fixed anyway.

What this fixes is that the _`$priority`_ parameter - which was the documented way to influence the run order of callbacks, wasn't being respected and now will be.

So if people want to use a specific run order of their callbacks, they should use the (documented) `$priority` parameter and not rely on registration order.

jrfnl: @schlessera See: https://3v4l.org/gAui9 We're not ordering the callbacks, we're ordering the priorities. That's a different level in the array.

schlessera (approved):

src/Hooks.php
schlessera: Yes, I understand your point now, and you're right: this is not affected by unstable sorting.

Issue #475, opened 2021-04-27
Improve testing instructions


schlessera:
[Testing instructions in the documentation](https://requests.ryanmccue.info/#testing) are lacking and fail to mention what the dependencies for some of the tests are and how to get all of them to be properly executed.

We should build a full `CONTRIBUTING.md` document with detailed instructions instead of this simplistic chapter.

PR #278, opened 2017-06-05
Add contributing doc

rmccue:
This lays out the philosophy behind the project, and how it should be maintained.

In addition to this, I'm adding more committers to the project: @dd32 and @ozh have both been contributing for a while; @robwilde and @ptasker both volunteered ([here](https://twitter.com/mrwilde/status/870627445589762048) and [here](https://twitter.com/petetasker/status/870631971986210816)) via Twitter to help out.

If any clarification is needed on these contribution notes, let me know; otherwise, any of you are welcome to review this PR, and merge it if accepted :)

PR #663, opened 2022-01-08
Try to fix GitHub Actions

simonhammes:
## Pull Request Type

- [X] I have checked there is no other PR open for the same change.

This is a:
- [ ] Bug fix
- [ ] New feature
- [X] Code quality improvement

## Detailed Description

The issue in https://github.com/WordPress/Requests/runs/4614423459?check_suite_focus=true might be related to https://github.com/composer/composer/issues/10389, which was fixed/released in https://github.com/composer/composer/releases/tag/2.2.4.

Unfortunately I could not get the tests to run locally, but the error message disappeared after downgrading to Composer 2.1.14.

I am actually trying to work on https://github.com/WordPress/Requests/issues/475 and maybe add a Dockerfile and a shell script to the repository to make it easier to run the tests locally.

## Quality assurance

- [ ] This change does NOT contain a breaking change (fix or feature that would cause existing functionality to change).

## Documentation

For new features:

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -33,6 +33,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
+          tools: composer:2.2.4
           coverage: none
 
       # At least one test needs a non-en_US locale to be available, so make sure it is.
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -33,7 +33,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          tools: composer:2.2.4
+          tools: composer:2.1.14
           coverage: none
 
       # At least one test needs a non-en_US locale to be available, so make sure it is.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
           php-version: ${{ matrix.php }}
           ini-values: error_reporting=-1, display_errors=On, log_errors_max_len=0
           coverage: ${{ steps.set_cov.outputs.COV }}
-          tools: cs2pr,composer:2.2.4
+          tools: cs2pr,composer:2.1.14
 
       # At least one test needs a non-en_US locale to be available, so make sure it is.
       - name: Install locales
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
           php-version: ${{ matrix.php }}
           ini-values: error_reporting=-1, display_errors=On, log_errors_max_len=0
           coverage: ${{ steps.set_cov.outputs.COV }}
-          tools: cs2pr
+          tools: cs2pr,composer:2.2.4
 
       # At least one test needs a non-en_US locale to be available, so make sure it is.
       - name: Install locales


simonhammes:
Yay, I got the tests passing using Composer 2.1.14.

It seems like Composer 2.2.4 did not fix the issue.

Another solution for this problem would be to modify https://github.com/RequestsPHP/test-server/blob/e9311dc36fa95e9f0f4077cdccbc8aede84f2199/bin/start.sh#L1 from `SERVERDIR="$PWD/$(dirname $0)"` to `SERVERDIR="$(dirname $0)"`.

Or maybe we could copy the necessary files from https://github.com/RequestsPHP/test-server into this repository to avoid such problems alltogether?

EDIT: Or apply a patch using https://github.com/cweagans/composer-patches.

jrfnl:
Hi @simonhammes, thank you for looking into this! 

I'd like to take a few steps back though and look at the various things you are saying one by one.

> I am actually trying to work on #475 and maybe add a Dockerfile and a shell script to the repository to make it easier to run the tests locally.

That's awesome! Please be aware though that Docker is no longer free for all, so may not be a sustainable solution. Also please be aware that people contributing to this library may be on a variety of OSes for their dev environment, so shell scripts would need to take that into account.

> Unfortunately I could not get the tests to run locally

Did you not get the tests to run at all locally or did most tests run with a select number of tests being skipped/failing ?
How are you trying to run the tests ?

> but the error message disappeared after downgrading to Composer 2.1.14.

While that may be a temporary work-around, it is not a solution and I would very much prefer a proper solution over a work-around which will bite us in the ass later on.

> EDIT: Or apply a patch using https://github.com/cweagans/composer-patches.

Same thing. I'd very much would like to avoid adding extra dependencies to work around a bug which should be fixed properly instead.

> The issue in https://github.com/WordPress/Requests/runs/4614423459?check_suite_focus=true might be related to composer/composer#10389, which was fixed/released in https://github.com/composer/composer/releases/tag/2.2.4.

> It seems like Composer 2.2.4 did not fix the issue.

Nice find! Composer 2.2 has turned out to be quite problematic all around so far, I've been involved in two other issues myself so far.

As that issue is closed, it may be worthwhile asking to reopen it it or to open a new, follow-up issue in the Composer repo outlining the problem you identified here.

> Another solution for this problem would be to modify https://github.com/RequestsPHP/test-server/blob/e9311dc36fa95e9f0f4077cdccbc8aede84f2199/bin/start.sh#L1 from `SERVERDIR="$PWD/$(dirname $0)"` to `SERVERDIR="$(dirname $0)"`.

I'm not sure that particular change would be the most stable cross-platform solution... ?

> Or maybe we could copy the necessary files from https://github.com/RequestsPHP/test-server into this repository to avoid such problems alltogether?

Actually, I think we should probably ask @rmccue if he's willing to give @schlessera and me admin access to the whole [RequestsPHP organization](https://github.com/RequestsPHP) as the test server repo is the only repo in it anyway.

Come to think of it, moving this repo to that organisation with Alain and me as (extra) admins, might have been the better choice instead of the move to the WP organisation last year....

Once we have maintainer access to the test server repo, giving the code there a good once-over and tech debt review would be great! That repo has barely been touched for the past five/six years, so there's bound to be some touching up needed. Having access to that repo and being able to maintain it, will also help with a few test cases for this repo which need to be added as it will allow us to generate additional server responses needed for those tests cases.

jrfnl:
@rmccue If you have a chance, it would be great to get your input on the above proposal for adding @schlessera and me to the Test server repo...

simonhammes:
> Hi @simonhammes, thank you for looking into this!

@jrfnl Thank you for getting back to me so quickly!

> > Unfortunately I could not get the tests to run locally
> 
> Did you not get the tests to run at all locally or did most tests run with a select number of tests being skipped/failing ? How are you trying to run the tests ?

Some were failing, although I got all of them passing now. That also allowed me to open #667.  
I will soon work on updating #278 with all relevant details on how to run the tests locally (without Docker or any extra tools).  
However, this requires Composer < 2.2 right now.

> While that may be a temporary work-around, it is not a solution and I would very much prefer a proper solution over a work-around which will bite us in the ass later on.

You are correct.

---

I will close this PR since Composer 2.2.5 did not fix the issue and we should find a stable solution to this problem.  
The comment in https://github.com/composer/composer/issues/10389#issuecomment-1007409643 also sounds like there won't be any more fixes concerning this problem.

rmccue:
> @rmccue If you have a chance, it would be great to get your input on the above proposal for adding @schlessera and me to the Test server repo...

@jrfnl Invited you and @schlessera as owners :)

FWIW: the test app is hosted on a deprecated version of the Heroku platform. It might be possible to move these to Docker containers (or GH Actions) instead, but it was intentionally cloud-hosted to ensure it goes through a full HTTP and networking stack, and that things like HTTP proxies are actually working as desired. I used Heroku specifically because it passed through the data without interfering, other platforms tended to mess with the input/output data/headers or fail when passed semi-invalid data.

(Happy to add you as collaborators on Heroku as well if desired, will need your Heroku account emails I think.)

jrfnl:
Thanks for that @rmccue ! I don't have a Heroku account myself. Maybe @schlessera has one ?

I have a niggly feeling that we're currently running the test app directly in GH Actions already, but would need to verify this to be sure.

rmccue:
Ah, I think _actually_ the issue was that they need to run behind a real server so that it exercises HTTPS tests against the real CA stack. We use the domain in the tests to set the constants:
https://github.com/WordPress/Requests/blob/73dbbbeb564d0bb290d49670c97731a735f464a7/tests/bootstrap.php#L14-L16

The test runner only overrides the HTTP version, not HTTPS:
https://github.com/WordPress/Requests/blob/73dbbbeb564d0bb290d49670c97731a735f464a7/.github/workflows/test.yml#L88

This host is only used for calls to `httpbin( ..., true )`, which (according to GH code search) is only used in `testHTTPS`: https://github.com/WordPress/Requests/blob/73dbbbeb564d0bb290d49670c97731a735f464a7/tests/Transport/BaseTestCase.php#L833

schlessera:
@rmccue We've fixed the code now. Would it be possible for you to add me to the Heroku app as a manager, so I can verify its state and push updates? I have a Heroku account under the email alain.schlesser@gmail.com.

rmccue:
@schlessera Added!

jrfnl:
Thanks @rmccue !

PR #669, opened 2022-02-02
GH Actions: add new workflow to auto-update certificate bundle

### LINKED ISSUE #635: Task: automate updating the certificates bundle
Thoughts:
* Use GH Actions with a daily cron schedule
* Check for changes to this page and/or RSS feed if available: https://curl.se/docs/caextract.html
* Use similar logic as used in the website update workflow to auto-generate a PR when a new bundle has been published.

jrfnl:
### Certificate: add checksum file

Add the checksum file for the certificate bundle to the package to allow both users of the package as well as maintainers of the package, to verify the validity of the included certificate bundle.

This file should only ever be updated at the same time as the certificate bundle is being updated.

### GH Actions: new workflow to automatically update certificate bundle 

This adds a new workflow which will automatically check the cURL website for an update to the certificate bundle once a day and if an updated bundle is found, it will automatically create a pull request against the `develop` branch to update the bundle in the Requests package.

The workflow will also update the certificate checksum file and verify the checksum of the downloaded certificate bundle.

Notes:
* A condition has been added to prevent the cron job from running on forks (to conserve resources).
* The workflow uses the recommended commands for automated downloads as per the https://curl.se/docs/caextract.html page.
    These recommended commands do a conditional download only when a file is changed and use an `etag*.txt` file to check whether the upstream file has changed.
    These `etag*.txt` files don't really need to be stored in the actual repo, so they have been added to the `.gitignore` file.
    In the workflow, these `etag*.txt` files are stored to and restored from a workflow cache to allow for the conditional download.
* While the workflow runs on a cron job and manual updating of the certificate file/checksum file should therefore never be needed, as an extra security measure, the workflow will also run whenever a PR is opened to update the certificate files or when a change to the certificate files is pushed to the `stable` or `develop` branch.
    Note: as PRs which are opened from within a workflow do not trigger new workflows to be run (= default behaviour for GitHub Actions), the PR potentially created by this workflow will not trigger a recursive run of this workflow.
* If a PR triggers this workflow and a certificate update would be needed, a (new) PR against the original PR will be opened with the certificate update.
* If the workflow is triggered via the cronjob or for a push against `stable`/`develop`, any PR which may be opened will be opened against `develop`.

Fixes #635


jrfnl:
Did some debugging and smoothed out the kinks.

PRs created due to the workflow being triggered via cronjob or due to a push to `stable`/`develop` will now open a pull request against `develop`.
PRs created due to a PR updating either the workflow or the certificate related files will be pulled against the PR branch which triggered the workflow.

schlessera (approved):

PR #679, opened 2022-02-07
`curl.after_send` hook: pass the `$handle` as parameter

### LINKED ISSUE #220: Missing Available Hook in Doc and Missing Hook Parameter

jrfnl:
Fixes #220

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -80,7 +80,7 @@ Available Hooks
 
     Run events after sending the request, but before interpreting the response.
 
-    Parameters: -
+    Parameters: `cURL resource|CurlHandle &$handle`
 
 * **`curl.after_request`**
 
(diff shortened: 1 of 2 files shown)


alpipego (approved):

jrfnl:
Before we actually action this, we should have a good look at the security implications of allowing access to the `$handle` (not just here).

schlessera:
The `curl.before_send` already provides access to the `$handle`, so it is questionable in the end whether this would have any security implications that aren't already present anyway.

So we need to consider the following:
* What would be an extra issue with the "after send" access that wouldn't already be problematic with the "before send"?
* Is the "before send" already problematic enough that we need to act on it?

PR #680, opened 2022-02-07
`*.after_request` hook: always pass `$info`

### LINKED ISSUE #321: $info parameter isn't passed to <transport>.after_request hook on non-blocking requests

jrfnl:
Callbacks hooked into the `[curl|fsockopen].after_request` hook would previously only receive the `$info` parameter for blocking requests. For non-blocking requests, the parameter would be omitted.

This made creating callbacks a little awkward, especially with both parameters being passed by reference.

This PR changes this to:
* Always pass the `$info` parameter by reference. When the request is non-blocking, this will just be a plain empty array.
* Update the hook documentation to reflect this change.

Note: this _could_ be considered a breaking change, so discussion may be warranted on whether this change can go into a minor or needs to wait for a major release.

Fixes #321

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -86,12 +86,14 @@ Available Hooks
 
     Alter the raw HTTP response before returning for parsing.
 
-    Parameters: `string &$headers`, `[array &$info]`
+    Parameters: `string &$headers`, `array &$info`
 
-    The optional `$info` parameter contains the associated array as defined in
+    The `$info` parameter contains the associated array as defined in
     the return value for [curl_getinfo()](https://www.php.net/curl-getinfo#refsect1-function.curl-getinfo-returnvalues).
 
-This optional parameter will be present when a blocking request was made (`$options['blocking' = true`) and will not be present when a non-blocking request was made (`$options['blocking' = false`). The callback signature needs to be adapted accordingly.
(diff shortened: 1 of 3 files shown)


alpipego (commented):
I'm fine with adding this to the next minor release. As far as I'm aware, there is no way of catching the `ArgumentCountError` anyway, so I don't see how it breaks existing code.

This may be out-of-scope for this PR, but maybe it's worth mentioning that the `$headers` string will also be empty for non-blocking requests?

--- 

I have confirmed the functional change in a WordPress context.

schlessera (approved):

jrfnl:
Thanks for reviewing @alpipego !

> This may be out-of-scope for this PR, but maybe it's worth mentioning that the `$headers` string will also be empty for non-blocking requests?

Good point. I've updated the commit to include that.

Issue #659, opened 2021-12-09
BC issue with mixed versions between WP Request and WP-CLI Request


schlessera:
## Summary

When WordPress uses strictly-typed arguments in constructors dealing with `Requests`, they break with `Requests` version mismatches, as the new objects do not inherit the old objects.

### Given the following code sample

As an example, let's consider installing WP Core with WP-CLI. This is only a single example for a bigger problem.

```bash
wp core install ...
```

### I'd expect the following behaviour

I'd expect this to work successfully, and WP-CLI doing what is needed to put WP Core into an installed state and then outputting:
```text
Success: WordPress installed successfully.
```

### Instead this happened

With Requests v2 bundled with WP-CLI, this fatals:

```php
thrown in /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-includes/class-wp-http-requests-response.php on line 42
Fatal error: Uncaught TypeError: WP_HTTP_Requests_Response::__construct(): Argument #1 ($response) must be of type Requests_Response, WpOrg\Requests\Response given, called in /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-includes/class-http.php on line 397 and defined in /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-includes/class-wp-http-requests-response.php:42
Stack trace:
#0 /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-includes/class-http.php(397): WP_HTTP_Requests_Response->__construct()
#1 /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-includes/class-http.php(626): WP_Http->request()
#2 /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-includes/http.php(162): WP_Http->get()
#3 /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-admin/includes/upgrade.php(522): wp_remote_get()
#4 /tmp/wp-cli-test-run--61afd4a9c43888.66717688/wp-admin/includes/upgrade.php(113): wp_install_maybe_enable_pretty_permalinks()
#5 vendor/wp-cli/core-command/src/Core_Command.php(618): wp_install()
#6 vendor/wp-cli/core-command/src/Core_Command.php(401): Core_Command->do_install()
#7 [internal function]: Core_Command->install()
#8 php/WP_CLI/Dispatcher/CommandFactory.php(100): call_user_func()
[...]

PR #681, opened 2022-02-09
Autoload: introduce method `get_deprecated_classes()`.

alpipego:
This method will return the array for the class mapping between the
legacy classes and their namespaced 2.x counterparts. The need for this
change came up while using the 2.x version of the library together
with a framework or CMS, that includes a 1.x version of Requests.

In order to prevent the framework from autoloading the 1.x classes,
they must be aliased before accessed. Through exposing the mapping
array, the classes can be aliased before being their implementations
get loaded.

See #659

## Pull Request Type

This is a:
- [ ] Bug fix
- [x] New feature

## Detailed Description

This adds the method `\WpOrg\Requests\Autoload::get_deprecated_classes` to access the class mapping from v1 to v2 and load the classes before `\WP_Http` gets the chance to do so.

## Quality assurance

## Documentation

For new features:

### DIFF
--- a/examples/preload-aliases.php
+++ b/examples/preload-aliases.php
@@ -1,0 +1,33 @@
+<?php
+
+/**
+ * If you need to run the Requests v2.x library with another framework/CMS that still ships v1.x,
+ * use the following to preload the class aliases before the actual implementations of Requests v1.x. get loaded.
+ *
+ * Please note that mixing Requests v1.x and v2.x files is a bad idea and may have side effects.
+ * Do so at your own risk.
+ */
+
+// First, include the Requests Autoloader.
+require_once 'path/to/Requests/src/Autoload.php';
+
+// Make sure the autoloader is registered first as otherwise you may run into trouble on PHP 8.1.
+// See: https://news-web.php.net/php.internals/115549
+WpOrg\Requests\Autoload::register();
+
+// Silence deprecations.
+// Depending on when the bootstrapping is done, you may need to wrap this in a `if (!defined(...)) {}`.
+define('REQUESTS_SILENCE_PSR0_DEPRECATIONS', true);
+
+$preload = WpOrg\Requests\Autoload::get_deprecated_classes();
+// Add the base Requests class, just to be safe.
+$preload['requests'] = '\WpOrg\Requests\Requests';
+
+// Preload the class aliases for the Requests 1.x classes to ensure only Requests 2.x classes get loaded.
+foreach ($preload as $old => $new) {
+	// Make sure we don't get "Class already exists errors" from autoloading chains
+	// Think: an `implements` causing an interface to be loaded before we explicitly request it.
+	if (class_exists($old) === false && interface_exists($old) === false) {
+		WpOrg\Requests\Autoload::load($old);
+	}
+}
--- a/examples/preload-aliases.php
+++ b/examples/preload-aliases.php
@@ -25,7 +25,7 @@ $preload['requests'] = '\WpOrg\Requests\Requests';
 
 // Preload the class aliases for the Requests 1.x classes to ensure only Requests 2.x classes get loaded.
 foreach ($preload as $old => $new) {
-	// Make sure we don't get "Class already exists errors" from autoloading chains
+	// Make sure we don't get "Class already exists errors" from autoloading chains.
 	// Think: an `implements` causing an interface to be loaded before we explicitly request it.
 	if (class_exists($old) === false && interface_exists($old) === false) {
 		WpOrg\Requests\Autoload::load($old);
--- a/src/Autoload.php
+++ b/src/Autoload.php
@@ -183,5 +183,14 @@ if (class_exists('WpOrg\Requests\Autoload') === false) {
 
 			return false;
 		}
+
+		/**
+		 * Get the array of deprecated Requests 1.x classes mapped to their equivalent Requests 2.x implementation.
+		 *
+		 * @return array
+		 */
+		public static function get_deprecated_classes() {
+			return self::$deprecated_classes;
+		}
 	}
 }
--- a/tests/AutoloadTest.php
+++ b/tests/AutoloadTest.php
@@ -5,6 +5,7 @@ namespace WpOrg\Requests\Tests;
 use Requests_Exception_Transport_cURL;
 use Requests_Utility_FilteredIterator;
 use WpOrg\Requests\Tests\TestCase;
+use WpOrg\Requests\Autoload;
 use WpOrg\Requests\Utility\FilteredIterator;
 
 final class AutoloadTest extends TestCase {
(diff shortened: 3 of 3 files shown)


jrfnl (commented):
@alpipego Looks like the example file mentioned in https://github.com/WordPress/Requests/issues/659#issuecomment-1033223918 is missing. Did you forget to push a second commit ?

alpipego:
> @alpipego Looks like the example file mentioned in [#659 (comment)](https://github.com/WordPress/Requests/issues/659#issuecomment-1033223918) is missing. Did you forget to push a second commit ?

Sorry, I amended my commit but didn't add the file. It's there now.

examples/autoload.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+// require Requests autoloader or use the one provided by composer.
```
jrfnl:
```suggestion
// First, include the Requests Autoloader.
```

The code would be different if using the Composer autoload file as that automatically registers the autoloader, so let's not confuse people.

examples/autoload.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
jrfnl: I think this code sample needs an introduction of when to use this type of bootstrapping (in a file docblock).

Should probably also include a warning that using a mix of Requests 1.x and 2.x files is a bad idea to start with and to do so at your own risk.

src/Autoload.php
```diff
@@ -183,5 +183,14 @@ public static function load($class_name) {
 
 			return false;
 		}
+
+		/**
+		 * Get the array of deprecated classes mapped to their equivalent implementation.
```
jrfnl:
```suggestion
		 * Get the array of deprecated Requests 1.x classes mapped to their equivalent Requests 2.x implementation.
```

jrfnl (commented):
Thanks @alpipego for getting this set up!

Generally looks good to me.

I do think the example file should be renamed though as this is not about the "normal" autoloading, but about _preloading_ the Requests 1.x class names.

examples/autoload.php
alpipego: I have changed this to your suggestion, do you also want me to change the `require_once` on the next line to a `include_once`?

I'm all for not confusing users, just an FYI: In my code I have preceded this with the Composer autoloader, and I'm not explicitly loading the Requests Autoloader.

alpipego:
Thank you for your notes, @jrfnl. 
I have amended my previous commit, I hope this is inline with your expectation. If you'd like me to add micro commits for changes within a PR, I'll gladly switch over to that.

examples/preload-aliases.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+/**
+ * If you need to run the Requests 2.x library with WordPress core that still ships v1.x,
+ * use the following to preload the class aliases before WordPress gets a chance to load
+ * the actual implementations of Requests 1.x.
```
jrfnl: I wonder if we should make this text a little bit more agnostic ? i.e. less specific to WordPress.
The same preloading could be needed in the context of other CMSes/frameworks.

jrfnl (commented):
@alpipego Thanks for the update! And yes, in this case, amending the commits is perfectly fine and makes for a clean commit history, so all good 👍🏻 

Aside from the inline remark, there is only one other thing:
What about adding a very simple test for the method. I understand if it feels redundant as there is no logic in the `Autoload::get_deprecated_classes()` method, but then again, there are also no other safeguards like a type declaration or code within the library calling the method.
So what about a simple test just calling the method and verifying that the returned value is an array and has # number of elements ?

examples/autoload.php
jrfnl: I think using `require_once` is appropriate here, so this should be fine.

examples/preload-aliases.php
jrfnl: Also, from my previous remark:

> Should probably also include a warning that using a mix of Requests 1.x and 2.x files is a bad idea to start with and to do so at your own risk.

alpipego:
I have amended my commit:
* I have added a test. Since [`\PHPUnit\Framework\Assert::assertCount`](https://phpunit.readthedocs.io/en/9.5/assertions.html#assertcount) checks for `Countable`/`is_iterable`, I have omitted testing for the `arary` type specifically. Please let me know if I should add it.
* I have reworded the commit message and example to be more framework-agnostic.

jrfnl (approved):
@alpipego Thanks for the update. Looks good to go to me! I'll give @schlessera a chance to look it over too, but as far as I'm concerned: let's merge this!

examples/preload-aliases.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+/**
+ * If you need to run the Requests v2.x library with another framework/CMS that still ships v1.x,
+ * use the following to preload the class aliases before the actual implementations of Requests v1.x. get loaded.
```
schlessera, line 5:
```suggestion
 * use the following to preload the class aliases before the actual implementations of Requests v1.x. gets loaded.
```

examples/preload-aliases.php
```diff
@@ -0,0 +1,33 @@
...
+// Depending on when the bootstrapping is done, you may need to wrap this in a `if (!defined(...)) {}`.
+define('REQUESTS_SILENCE_PSR0_DEPRECATIONS', true);
+
+$preload = WpOrg\Requests\Autoload::get_deprecated_classes();
+// Add the base Requests class, just to be safe.
+$preload['requests'] = '\WpOrg\Requests\Requests';
+
+// Preload the class aliases for the Requests 1.x classes to ensure only Requests 2.x classes get loaded.
+foreach ($preload as $old => $new) {
+	// Make sure we don't get "Class already exists errors" from autoloading chains
```
schlessera:
```suggestion
	// Make sure we don't get "Class already exists errors" from autoloading chains.
```

src/Autoload.php
```diff
@@ -183,5 +183,14 @@ public static function load($class_name) {
...
 		}
+
+		/**
+		 * Get the array of deprecated Requests 1.x classes mapped to their equivalent Requests 2.x implementation.
+		 *
+		 * @return array
+		 */
+		public static function get_deprecated_classes() {
+			return self::$deprecated_classes;
+		}
```
schlessera, line 194: This is fine for me, but I was wondering whether it might be easier here to just turn the `private static $deprecated_classes` into a (public by default) `const DEPRECATED_CLASSES` instead. This would remove the need for an additional method.

schlessera (changes requested):

examples/preload-aliases.php
alpipego: If the verb is singular, the noun should then also be, right; i.e. `the actual implementation … gets`?

src/Autoload.php
jrfnl: I seem to remember we discussed that before and it would cause a PHP 5.6 error as we're checking against the array with `isset( $array['key'] )` and that syntax is not supported for PHP 5.6 with class constants.

So, having the method is the better solution. It will not prevent us from changing the property to a constant once the minimum supported PHP version goes up to PHP 7.0 (or higher), but will also not necessitate a less efficient/slower implementation of the autoloader.

examples/preload-aliases.php
jrfnl: Correct.

schlessera: Oh, right, I misead.

schlessera (approved):

PR #689, opened 2022-02-28
Refine docs for `curl.before_send`

alpipego:
Change "just before" to "before". It is more inline with the doc for
`fsockopen.before_send` and more correct.


jrfnl (approved):
Thanks @alpipego !

PR #695, opened 2022-03-10
Docs: various improvements [2]

jrfnl:
Next PR in a series of documentation style & completeness improvements started with #685

By no means claiming completeness, but this is at least another step forward.

* Add missing parameter tags
* Add missing param comments
* Add missing property docblock
* Add missing `@throws` comment
* Add missing function docblocks
    Notably, remove a couple of `@inheritDoc`s for methods where the PHP native methods which are being overloaded don't have a notable docblock, so there is nothing to inherit.

### DIFF
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -36,8 +36,8 @@ class Cookie {
 	/**
 	 * Cookie attributes
 	 *
-	 * Valid keys are (currently) path, domain, expires, max-age, secure and
-	 * httponly.
+	 * Valid keys are `'path'`, `'domain'`, `'expires'`, `'max-age'`, `'secure'` and
+	 * `'httponly'`.
 	 *
 	 * @var \WpOrg\Requests\Utility\CaseInsensitiveDictionary|array Array-like object
 	 */
@@ -46,8 +46,7 @@ class Cookie {
 	/**
 	 * Cookie flags
 	 *
-	 * Valid keys are (currently) creation, last-access, persistent and
-	 * host-only.
+	 * Valid keys are `'creation'`, `'last-access'`, `'persistent'` and `'host-only'`.
 	 *
 	 * @var array
 	 */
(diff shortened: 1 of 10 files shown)


src/Cookie.php
```diff
@@ -66,11 +66,13 @@ class Cookie {
...
-	 * @param string $name
-	 * @param string $value
+	 * @param string                                                  $name           The name of the cookie.
+	 * @param string                                                  $value          The value for the cookie.
 	 * @param array|\WpOrg\Requests\Utility\CaseInsensitiveDictionary $attributes Associative array of attribute data
-	 * @param array $flags
-	 * @param int|null $reference_time
+	 * @param array                                                   $flags          The flags for the cookie.
+	 *                                                                                Valid keys are (currently) creation,
+	 *                                                                                last-access, persistent and host-only.
```
schlessera: Not sure about the `(currently)` here. Isn't all of the source code - including the docblocks - just the current version of the library?

src/Transport/Curl.php
```diff
@@ -467,7 +467,7 @@ private function setup_handle($url, $headers, $data, $options) {
 	 * @param string $response Response data from the body
 	 * @param array $options Request options
 	 * @return string|false HTTP response data including headers. False if non-blocking.
-	 * @throws \WpOrg\Requests\Exception
+	 * @throws \WpOrg\Requests\Exception If the request resulted in a Curl error.
```
schlessera: This probably shouldn't be the PSR-4 class notation, but the actual name of the project:
```suggestion
	 * @throws \WpOrg\Requests\Exception If the request resulted in a cURL error.
```

schlessera (dismissed):

src/Cookie.php
jrfnl: True. To be fair, I copied this text (including the `(currently)` from the docblock for the property to which the `$flags` parameter will be assigned, so I think I should probably removed it from both in that case.

schlessera (approved):

PR #698, opened 2022-03-21
GH Actions: version update for various action runners

jrfnl:
### GH Actions: version update for actions/cache

The update doesn't actually contain any changed functionality, it's mostly just a change of the Node version used by the action itself (from Node 12 to Node 16).

Also saw that I missed updating the `actions/checkout` for this new workflow and in some other places. Ref: #691

Refs:
* https://github.com/actions/cache/releases/

### 🆕 GH Actions: version update for peter-evans/create-pull-request

The update contain minimal changed functionality, it's mostly just a change of the Node version used by the action itself (from Node 12 to Node 16).

There is one breaking change regarding the `add-paths` key, but as that's not used in our workflows this doesn't affect us.

Refs:
* https://github.com/peter-evans/create-pull-request/releases/

### DIFF
--- a/.github/workflows/cs.yml
+++ b/.github/workflows/cs.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install PHP
         uses: shivammathur/setup-php@v2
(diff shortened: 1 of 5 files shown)


jrfnl:
I've added a second commit to this PR for more of the same. See the updated PR description.

schlessera (approved):

PR #706, opened 2022-04-02
GH Actions: show deprecations when linting

jrfnl:
While rare, there are some deprecations which PHP can show when a file is being linted.
By default these are ignored by PHP-Parallel-Lint.

Apparently though, there is an option to show them (wasn't documented until recently), so let's turn that option on.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
   },
   "scripts": {
     "lint": [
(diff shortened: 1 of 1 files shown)


schlessera (approved):

schlessera:
Nice one, didn't know about this flag. 👍

jrfnl:
Neither did I until I saw a suggestion from someone to document it 😂 (which I promptly did).

Issue #593, opened 2021-11-06
Add input validation


jrfnl:
This is just an issue to track which classes have been reviewed for missing input validation and the associated PRs adding the input validation.

Input validation is being added to all entry point methods for each class with an `InvalidArgument` exception being thrown for invalid parameter types. As this _may_ break existing uses of Requests, this input validation needs to be added in the 2.0.0 release.

### Action list
- [x] `WpOrg\Requests\Auth\Basic` - PR #574
- [x] `WpOrg\Requests\Autoload` - No additional input validation needed, the `Autoload::load()` method should only ever by called by PHP itself.
- [x] `WpOrg\Requests\Cookie` - PR #609
- [x] `WpOrg\Requests\Cookie\Jar` - PR #610
- [x] `WpOrg\Requests\Hooks` - PR #573
- [x] `WpOrg\Requests\IdnaEncoder` - PR #592 (only typical entry point, other methods deemed unnecessary)
- [x] `WpOrg\Requests\Ipv6` - PR #601
- [x] `WpOrg\Requests\Iri` - PR #602
- [x] `WpOrg\Requests\Port` - Input validation was included when the class was introduced in PR #538
- [x] `WpOrg\Requests\Proxy\Http` - PR #611
- [x] `WpOrg\Requests\Requests` - PR #621
- [x] `WpOrg\Requests\Response` - PR #603
- [x] `WpOrg\Requests\Response\Headers` - PR #605
- [x] `WpOrg\Requests\Session` - PR #620
- [x] `WpOrg\Requests\Ssl` - PR #572
- [x] `WpOrg\Requests\Transport\Curl` - PR #629
- [x] `WpOrg\Requests\Transport\Fsockopen` - PR #629
- [x] `WpOrg\Requests\Utility\CaseInsensitiveDictionary` - No changes needed.
        - The `__construct()` method already has a type declared for the parameter.
        - The `offset*()` methods are part of `ArrayAccess()` and they should not be called directly, but only indirectly and when called that way, will receive the correct input type.
        - The remaining methods don't take parameters.
- [x] `WpOrg\Requests\Utility\FilteredIterator` - PR #604
- [x] `WpOrg\Requests\Utility\InputValidator` - Not applicable.
- [x] `WpOrg\Requests\Exception` - I'm going to leave this as-is. The important parameters are validated via the PHP native `Exception` class, that should be enough.
- [x] `WpOrg\Requests\Exception\ArgumentCount` - Only intended for internal use.
[...]

PR #710, opened 2022-04-06
Tests: expand type verification tests

jrfnl:
This PR is a follow-up on the PRs created to address #593.

At the time, it was discussed in a call that it would be better to always test against all types which would generate an exception, but with the pressure to release version 2.0.0, this wasn't implemented at that time.

This PR addresses this remark now.

### Tests: add new TypeProviderHelper class

... which can be used as a data provider for all tests which need to test against a variation of different types of data.

Typical groups of keys are defined as class constants for use as selectors.

As a resource is created for the test cases, the `clean_up()` method needs to be called after the tests using data from the TypeProviderHelper have run.
As the majority of the test classes will use the TypeProviderHelper for one or more tests, we may as well generically call this clean up method from the base `TestCase`.

Includes adding an exception for the use of camelCase method names to the PHPCS ruleset.
The `WordPress.NamingConventions.ValidFunctionName` sniff already doesn't trigger on the use of camelCase in the test classes themselves (as they extend another class), so let's blanket ignore the snake_case method name rule for all files in the `tests` directory.
Note: if so preferred, I could add a sniff to enforce using camelCase in the tests directory, though I'd need to see if I can find one which could make an exception for the snake_case fixture methods.

### InputValidatorTest: refactor to use the TypeProviderHelper

This removes the original data provider in favour of having a "valid" and "invalid" test function for each function with a dedicated data provider for each.

The refactor is nearly 1-on-1: all previously tested cases are still tested, with the one exception being that a curl handle is not included in the "default" data provided via the TypeProviderHelper.

### Tests: implement use of the new TypeProviderHelper class

.. in all the data providers related to tests which check for input types.

The benefit of this refactor is that if/when additional test cases are added to the `TypeProviderHelper`, these tests will automatically also be tested against those new test cases.

### DIFF
--- a/tests/TypeProviderHelper.php
+++ b/tests/TypeProviderHelper.php
@@ -1,0 +1,315 @@
+<?php
+
+namespace WpOrg\Requests\Tests;
+
+use ArrayIterator;
+use EmptyIterator;
+use stdClass;
+use WpOrg\Requests\Tests\Fixtures\ArrayAccessibleObject;
+use WpOrg\Requests\Tests\Fixtures\StringableObject;
+
+/**
+ * Helper class to provide an exhaustive list of types to test type safety.
+ */
+final class TypeProviderHelper {
+
+	/**
+	 * Keys of all type entries representing null.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_NULL = ['null'];
+
+	/**
+	 * Keys of all type entries representing a boolean.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_BOOL = [
+		'boolean false',
+		'boolean true',
+	];
+
+	/**
+	 * Keys of all type entries representing an integer.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_INT = [
+		'integer 0',
+		'negative integer',
+		'positive integer',
+	];
+
+	/**
+	 * Keys of all type entries representing a float.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_FLOAT = [
+		'float 0.0',
+		'negative float',
+		'positive float',
+	];
+
+	/**
+	 * Keys of all type entries representing an integer or float.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_INT_FLOAT = [
+		'integer 0',
+		'negative integer',
+		'positive integer',
+		'float 0.0',
+		'negative float',
+		'positive float',
+	];
+
+	/**
+	 * Keys of all type entries representing a string.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_STRING = [
+		'empty string',
+		'numeric string',
+		'textual string',
+		'textual string starting with numbers',
+	];
+
+	/**
+	 * Keys of all type entries which are stringable.
+	 *
+	 * @var string[]
+	 */
+	const GROUP_STRINGABLE = [
+		'empty string',
+		'numeric string',
+		'textual string',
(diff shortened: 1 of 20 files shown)


tests/TypeProviderHelper.php
```diff
@@ -0,0 +1,241 @@
...
+
+use ArrayIterator;
+use EmptyIterator;
+use stdClass;
+use WpOrg\Requests\Tests\Fixtures\ArrayAccessibleObject;
+use WpOrg\Requests\Tests\Fixtures\StringableObject;
+
+final class TypeProviderHelper {
+
+	/** Keys of all type entries representing null. */
```
schlessera: We should add type information for these constants as well.
```suggestion
	/**
	 * Keys of all type entries representing null.
	 *
	 * @var array<string>
	 */
```

tests/TypeProviderHelper.php
```diff
@@ -0,0 +1,241 @@
...
+	public static function getAllExcept(array ...$except) {
+		$except = array_flip(array_merge(...$except));
+
+		return array_diff_key(self::getAll(), $except);
+	}
+
+	/**
+	 * Retrieve an array in data provider format with a selection of typical PHP data types.
+	 *
+	 * @param array ...$selection One or more array containing the names of the types to include.
```
schlessera:
```suggestion
	 * @param array ...$selection One or more arrays containing the names of the types to include.
```

tests/TypeProviderHelper.php
```diff
@@ -0,0 +1,241 @@
...
+	/**
+	 * Retrieve an array in data provider format with a selection of all typical PHP data types
+	 * *except* the named types specified in the $except parameter.
+	 *
+	 * @param array ...$except One or more arrays containing the names of the types to exclude.
+	 *                         Typically, one or more of the predefined "groups" (see the constants)
+	 *                         would be used here.
+	 *
+	 * @return array
+	 */
```
schlessera: Should we be precise about the structure of the arrays everywhere?
```suggestion
	/**
	 * Retrieve an array in data provider format with a selection of all typical PHP data types
	 * *except* the named types specified in the $except parameter.
	 *
	 * @param array<string> ...$except One or more arrays containing the names of the types to exclude.
	 *                                 Typically, one or more of the predefined "groups" (see the constants)
	 *                                 would be used here.
	 *
	 * @return array<string>
	 */
```

tests/TypeProviderHelper.php
```diff
@@ -0,0 +1,241 @@
...
+
+namespace WpOrg\Requests\Tests;
+
+use ArrayIterator;
+use EmptyIterator;
+use stdClass;
+use WpOrg\Requests\Tests\Fixtures\ArrayAccessibleObject;
+use WpOrg\Requests\Tests\Fixtures\StringableObject;
+
+final class TypeProviderHelper {
```
schlessera, line 14: This could use a piece of documentation, I guess.

Something like this, maybe?
```suggestion
/**
 * Helper class to provide an exhaustive list of types to test against type safety.
 */
final class TypeProviderHelper {
```

schlessera (changes requested):

tests/TypeProviderHelper.php
jrfnl: For this and the other remarks: absolutely valid points, I was just cheating/taking a shortcut as this is test only code, but you're right, let's lead by example and do it properly.

schlessera (approved):

Issue #648, opened 2021-11-25
Tests: re-organize and split large files


jrfnl:
What with the expansion of the test suite over the last few months, a number of test files are becoming quite large and it is not always straight forward anymore to navigate/find what's tested and where.

I'd like to propose to re-organize the test suite along the following lines:
* Each class translates to a subdirectory in the `tests` directory.
    Classes in subdirectories in `src` will follow their namespace path first and then have a subdirectory for the class itself.
    Examples:
    - The tests for the `WpOrg\Requests\Response` class would live in the `tests/Response` subdirectory.
    - The tests for the `WpOrg\Requests\Utility\FilteredIterator` class would live in the `tests/Utilities/FilteredIterator` subdirectory.
* Tests which are testing a class as a whole are put in an `ClassNameTest` class in this (sub)directory.
* Unit tests for specific methods each get their own test class and file, where the class and file are named after the method being tested.
    For instance, tests specifically for the `WpOrg\Requests\Cookie::domain_matches()` method would live in a `WpOrg\Requests\Tests\Cookie\DomainMatchesTest` class.
    For closely linked groups of methods, it is perfectly fine to use one test class, as long as the name of the test class is descriptive.
    Example: tests for array access methods implemented in a class, could be placed in one test class, with the class being called `ArrayAccessTest`.

When re-organizing the tests along the above lines, I'd also like to propose to order tests in the new classes along the following lines:
* Tests are ordered by the logic order within the method under test.
    For example: if a method under test roughly does: 1) input validation, 2) something conditional 3) something unconditional.
    The tests within the class would be ordered along the same lines, i.e. first the tests for the input validation, then the tests covering the conditional action, then the tests covering the unconditional action.
* Data providers are closely linked to the tests they are used by in two ways:
    1. If the data provider is only used by one test, the name of the data provider should mirror the name of the test method, replacing `test`with `data`, i.e. the `testSomething()` method would have a data provider called `dataSomething()`.
    2. Method order: the data provider will directly follow the test it is used for. If a data provider is used by multiple test methods, it will follow the last of these methods.
* Test "helper" methods will be placed at the end of a test class and should be named descriptively with regards to their function in the tests.

@schlessera This is roughly what we discussed in one of our calls. Did I miss anything ? Anything else we should consider/create guidelines for before starting this action ?

jrfnl:
@schlessera Mind looking this over and leaving an opinion ? If we are agreed, I can start getting this set up.

[...]

PR #727, opened 2022-04-25
SslTest: split up test class

jrfnl:
### Tests: introduce an SslTestCase class

... containing various methods which are used by tests for multiple different methods.

### SslTest: rename to VerifyCertificateTest

... as the remaining tests relate directly to the `verify_certificate()` method.

Including re-ordering the methods in the class to follow the order of the code in the method under test.

Related to #648

### DIFF
--- a/tests/Ssl/SslTestCase.php
+++ b/tests/Ssl/SslTestCase.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace WpOrg\Requests\Tests\Ssl;
+
+use WpOrg\Requests\Tests\Fixtures\StringableObject;
+use WpOrg\Requests\Tests\TestCase;
+
+abstract class SslTestCase extends TestCase {
+
+	/**
+	 * Test helper to mock a certificate.
+	 *
+	 * @param string      $dnsname  DNS name to match against.
+	 * @param bool|string $with_san Optional. How to generate the fake certificate.
+	 *                              - false:  plain, CN only;
+	 *                              - true:   CN + subjectAltName, alt set to same as CN;
+	 *                              - string: CN + subjectAltName, alt set to string value.
+	 *
+	 * @return array
+	 */
+	protected function fakeCertificate($dnsname, $with_san = true) {
+		$certificate = [
+			'subject' => [
+				'CN' => $dnsname,
+			],
+		];
+
+		if ($with_san !== false) {
+			// If SAN is set to true, default it to the dNSName
+			if ($with_san === true) {
+				$with_san = 'DNS: ' . $dnsname;
+			}
+
+			$certificate['extensions'] = [
+				'subjectAltName' => $with_san,
+			];
+		}
+
+		return $certificate;
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @return array
+	 */
+	public function dataMatch() {
+		return [
+			'top-level domain (stringable object)' => [
+				'host'      => new StringableObject('example.com'),
+				'reference' => new StringableObject('example.com'),
+			],
+			'subdomain' => [
+				'host'      => 'test.example.com',
+				'reference' => 'test.example.com',
+			],
+			'subdomain with wildcard reference' => [
+				'host'      => 'test.example.com',
+				'reference' => '*.example.com',
+			],
+		];
+	}
+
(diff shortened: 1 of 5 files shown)


tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+	 */
+	protected function fakeCertificate($dnsname, $with_san = true) {
+		$certificate = [
+			'subject' => [
+				'CN' => $dnsname,
+			],
+		];
+
+		if ($with_san !== false) {
+			// If SAN is set to true, default it to the dNSName
```
schlessera:
```suggestion
			// If SAN is set to true, default it to the dNSName.
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @return array
+	 */
+	public function dataNoMatch() {
+		return [
+			// Check that we need at least 3 components
```
schlessera:
```suggestion
			// Check that we need at least 3 components.
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+			'not a domain; wildcard reference' => [
+				'host'      => 'com',
+				'reference' => '*',
+			],
+			'domain name; wildcard on TLD as reference' => [
+				'host'      => 'example.com',
+				'reference' => '*.com',
+			],
+
+			// Check that double wildcards don't work
```
schlessera:
```suggestion
			// Check that double wildcards don't work.
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+			'domain name; double wildcard in reference' => [
+				'host'      => 'abc.def.example.com',
+				'reference' => '*.*.example.com',
+			],
+
+			// Check that we only match with the correct number of components
```
schlessera:
```suggestion
			// Check that we only match with the correct number of components.
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+			'four-level domain; three-level reference' => [
+				'host'      => 'abc.def.example.com',
+				'reference' => 'def.example.com',
+			],
+			'four-level domain; three-level wildcard reference' => [
+				'host'      => 'abc.def.example.com',
+				'reference' => '*.example.com',
+			],
+
+			// Check that the wildcard only works as the full first component
```
schlessera:
```suggestion
			// Check that the wildcard only works as the full first component.
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+			'four-level domain; four-level reference, but wildcard not stand-alone' => [
+				'host'      => 'abc.def.example.com',
+				'reference' => 'a*.def.example.com',
+			],
+
+			// Check that wildcards are not allowed for IPs
```
schlessera:
```suggestion
			// Check that wildcards are not allowed for IPs.
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+			'IP address; wildcard in refence (start)' => [
```
schlessera:
```suggestion
			'IP address; wildcard in reference (start)' => [
```

tests/Ssl/SslTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+				'host'      => '192.168.0.1',
+				'reference' => '*.168.0.1',
+			],
+			'IP address; wildcard in refence (end)' => [
```
schlessera:
```suggestion
			'IP address; wildcard in reference (end)' => [
```

schlessera (changes requested):

jrfnl:
@schlessera Happy to address these things, but they are all outside the scope of this PR, which just splits the test class up into method based classes, without making any other changes....

schlessera (approved):

PR #740, opened 2022-05-10
Tests: new `Cookie\FormatTest` class

jrfnl:
... with dedicated tests for the `Cookie::__toString()`, `Cookie::format_for_header()` and the `Cookie::format_for_set_cookie()` methods.

⚠️ Also take note that, IMO, some of these tests highlight potential bugs. I'd like a second opinion on this though before addressing this.

### DIFF
--- a/tests/Cookie/CookieTest.php
+++ b/tests/Cookie/CookieTest.php
@@ -12,21 +12,6 @@ final class CookieTest extends TestCase {
 
 		$this->assertSame('requests-testcookie', $cookie->name);
 		$this->assertSame('testvalue', $cookie->value);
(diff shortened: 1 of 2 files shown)


schlessera:
> Also take note that, IMO, some of these tests highlight potential bugs. I'd like a second opinion on this though before addressing this.

Should we discuss this here or in a separate issue to unblock and merge this PR ?

jrfnl:
I've opened #743 to discuss the potential bugs.

jrfnl:
I've rebased the PR after the recent merges and added one additional test case to the `dataFormat()` data provider - "Has key, empty value".

schlessera (approved):

PR #746, opened 2022-05-24
Tests: add dedicated tests for the Response::throw_for_status() method

jrfnl:


tests/Response/ThrowForStatusTest.php
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Verify that a redirection does not lead to an exception when redirection is allowed.
+	 *
+	 * @doesNotPerformAssertions
+	 */
+	public function testRedirectAndAllowed() {
+		$response              = new Response();
+		$response->status_code = 302;
+
+		$response->throw_for_status();
```
schlessera, line 25: Testing against a "default value" here seems like an potentially problematic assumption in the tests.

For better reliability, I would suggest:
- always using explicit values for each call
- adding one extra test for testing the current default value

jrfnl: Done (fixed up in the original commit).

schlessera (approved):

PR #752, opened 2022-06-20
Composer: raise the PHP minimum to PHP 5.6.20

jrfnl:
... which is in line with WordPress Core and prevents us potentially having to deal with bugs in PHP 5.6.0 to 5.6.19.

Includes updating the relevant documentation (hope I caught all references).

### DIFF
--- a/docs/goals.md
+++ b/docs/goals.md
@@ -18,7 +18,7 @@ Goals
 
    No matter what you have installed on your system, you should be able to run
    Requests. We use cURL if it's available, and fallback to sockets otherwise.
-   We require only a baseline of PHP 5.6, leaving the choice of PHP minimum
+   We require only a baseline of PHP 5.6.20, leaving the choice of PHP minimum
    requirement fully in your hands, and giving you the ability to support many
    more hosts.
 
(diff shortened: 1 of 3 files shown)


docs/goals.md
```diff
@@ -18,7 +18,7 @@ Goals
 
    No matter what you have installed on your system, you should be able to run
    Requests. We use cURL if it's available, and fallback to sockets otherwise.
-   We require only a baseline of PHP 5.6, leaving the choice of PHP minimum
+   We require only a baseline of PHP 5.6.20, leaving the choice of PHP minimum
    requirement fully in your hands, and giving you the ability to support many
    more hosts.
```
schlessera, line 23: Good for this change now, but I think this sentence needs rephrasing, regarding the "baseline", which doesn't mean much, and we're actually NOT "leaving the choice of PHP minimum requirement fully in your hands"...

schlessera (approved):

docs/goals.md
jrfnl: IIRC this is part of the "in comparison to other HTTP frameworks", highlighting the _low_ PHP version Requests still supports in contrast to other HTTP frameworks, which raise the minimum PHP version far more frequently and aggressively.

PR #754, opened 2022-06-20
Cookie\DomainMatchesTest: improve and stabilize

jrfnl:
### Cookie\DomainMatchesTest::testDomainMatch*(): split type check tests off to separate test method

... with a `TypeProviderHelper` based data provider to safeguard all invalid types will be rejected.

### Cookie\DomainMatchesTest::testDomainMatch*(): use named data provider

This commit:
* Adds descriptive names to each of the cases being tested via the data provider.
* Adds keys to the parameters passed for each test case to make it easier to understand the test case.
* Improves the punctuation for the inline documentation in the data provider.

### Cookie\DomainMatchesTest::testDomainMatch*(): add extra test cases

.. to safeguard the existing behaviour.

### Cookie::domain_matches(): minor code simplifications [1]

Remove a few redundant calculations and a redundant function call.

### Cookie::domain_matches(): minor code simplifications [2]

Store the value of the repeated strlen calculation for re-use instead of calling the function multiple times.

### DIFF
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -207,12 +207,12 @@ class Cookie {
 			return false;
 		}
 
-		if (substr($domain, -1 * strlen($cookie_domain)) !== $cookie_domain) {
+		if (substr($domain, -(strlen($cookie_domain))) !== $cookie_domain) {
 			// The cookie domain should be a suffix of the passed domain.
 			return false;
 		}
 
-		$prefix = substr($domain, 0, strlen($domain) - strlen($cookie_domain));
+		$prefix = substr($domain, 0, -strlen($cookie_domain));
 		if (substr($prefix, -1) !== '.') {
 			// The last character of the passed domain that is not included in the
 			// domain string should be a %x2E (".") character.
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -201,18 +201,19 @@ class Cookie {
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


src/Cookie.php
```diff
@@ -207,12 +207,12 @@ public function domain_matches($domain) {
...
 		}
 
-		if (substr($domain, -1 * strlen($cookie_domain)) !== $cookie_domain) {
+		if (substr($domain, -(strlen($cookie_domain))) !== $cookie_domain) {
 			// The cookie domain should be a suffix of the passed domain.
 			return false;
 		}
 
-		$prefix = substr($domain, 0, strlen($domain) - strlen($cookie_domain));
+		$prefix = substr($domain, 0, -strlen($cookie_domain));
```
schlessera: Why is one occurrence of `strlen()` being put into parentheses while the other isn't?

Also, it might be good to calculate the `$prefix` only once.

schlessera (changes requested):

src/Cookie.php
schlessera: NVM the second remark, it's not the same `substr()`.

jrfnl: I've added an extra commit which stores the calculated value and then automatically gets rid of the extra set of parentheses as well.

schlessera (approved):

PR #756, opened 2022-06-20
Cookie\ConstructorTest: add additional tests

jrfnl:
... to ensure the method is fully covered.

### DIFF
--- a/tests/Cookie/ConstructorTest.php
+++ b/tests/Cookie/ConstructorTest.php
@@ -6,6 +6,7 @@ use WpOrg\Requests\Cookie;
 use WpOrg\Requests\Exception\InvalidArgument;
 use WpOrg\Requests\Tests\TestCase;
 use WpOrg\Requests\Tests\TypeProviderHelper;
+use WpOrg\Requests\Utility\CaseInsensitiveDictionary;
 
 /**
  * @covers \WpOrg\Requests\Cookie::__construct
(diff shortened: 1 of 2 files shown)


tests/Cookie/ConstructorTest.php
```diff
@@ -128,4 +129,236 @@ public function testInvalidReferenceTime($input) {
...
+				'flags'    => [],
+				'expected' => [
+					'persistent' => false,
+					'host-only'  => true,
+				],
+			],
+			'all supported flags passed' => [
+				'flags'    => [
+					'creation'    => 543547575,
+					'last-access' => 543543555,
```
schlessera: Looks like `creation` date comes after `last-access` date here.

tests/Cookie/ConstructorTest.php
```diff
@@ -128,4 +129,236 @@ public function testInvalidReferenceTime($input) {
...
+					'persistent'  => true,
+					'host-only'   => false,
+				],
+				'expected' => [
+					'creation'    => 543547575,
+					'last-access' => 543543555,
```
schlessera: Looks like `creation` date comes after `last-access` date here.

schlessera (changes requested):

schlessera (approved):

tests/Cookie/ConstructorTest.php
jrfnl: True - no validation is done on the _value_ of these, though not sure if adding validation would add anything.

In the mean time, I've just fixed up the date for now ;-)

PR #760, opened 2022-07-04
Cookie\ParseTest: improve tests

jrfnl:
### Cookie\ParseTest::check_parsed_cookie(): various improvements

* Stabilize assertions by ensuring that we have a Cookie object to begin with.
* Ensure all assertions have a descriptive `$message` parameter.
* Improve documentation docblock.

### Cookie\ParseTest: add new test which explicitly tests the name/value parsing

While the name/value parsing was previously already tested via the `testParsingHeader()` with the two "Basic parsing" test cases a number of combination of situations were not being safeguarded, i.e.:
* How does the `Cookie::parse()` method handle when the cookie contains a key and an explicit name is also passed ?
* How is leading/trailing whitespace around keys/values handled ?

Each of these situations is now explicitly tested and the function behaviour documented.

Includes making the `$expected_attributes` parameter of the `check_parsed_cookie()` helper method optional.

Includes removing some test code from the `CookieTest` class which is now covered by the tests in this method.

### Cookie\ParseTest::testParsingHeader*(): use named data provider

This commit:
* Adds descriptive names to each of the cases being tested via the `dataParseResult()` data provider.
* Adds keys to the parameters passed for each test case to make it easier to understand the test case.

### Cookie\ParseTest::testParsingHeader*(): add extra test cases

... for functionality not yet safeguarded by the existing tests in this class.

### Cookie\ParseTest::testParsingHeader*(): remove extraneous parameter

The `dataParseResult()` dataprovider does not contain any test cases with `$flag` expectations.

### Cookie\ParseTest::testParsingHeaderWithOrigin(): use named data provider

This commit:
* Adds descriptive names to each of the cases being tested via the `dataParsingHeaderWithOrigin()` data provider.
* Adds keys to the parameters passed for each test case to make it easier to understand the test case.
* Don't use Perl-style (hash) comments.

### Cookie\ParseTest::dataParsingHeaderWithOrigin(): make invalid expectation more descriptive

While the test method only checks if the key is set and bases what assertion to run on that, it is still more descriptive to make it clear that this test case is expected to be regarded as invalid by setting the value to `true`.

### Cookie\ParseTest::testParsingHeaderWithOrigin(): add two extra test cases

... with additional origin variations to cover additional code branches/paths.

### Cookie\ParseTest: add new test which explicitly tests empty Headers handling

... to raise code coverage.

### Cookie\ParseTest: add new test which explicitly tests parsing of multi-cookie Headers

... to safeguard that multi-cookie Headers are handled correctly.

### Cookie\ParseTest: add appropriate `@covers` tags to remaining tests

... which didn't have a `@covers` tag yet (or which had one which was unnecessary).

### DIFF
--- a/tests/Cookie/ParseTest.php
+++ b/tests/Cookie/ParseTest.php
@@ -334,14 +334,25 @@ final class ParseTest extends TestCase {
 	}
 
 	/**
+	 * Verify parsing of cookies in Header objects when origin is known.
+	 *
 	 * @dataProvider dataParsingHeaderWithOrigin
+	 *
+	 * @param string $header              Cookie header string.
+	 * @param string $origin              URI for comparing cookie origins.
+	 * @param array  $expected            Array with expectations to be verified via check_parsed_cookie().
+	 *                                    Keys which can be set to be verified: 'name', 'value', 'expired'.
+	 * @param array  $expected_attributes Attribute expectations to verify.
+	 * @param array  $expected_flags      Flag expectations to verify.
+	 *
+	 * @return void
 	 */
 	public function testParsingHeaderWithOrigin($header, $origin, $expected, $expected_attributes = [], $expected_flags = []) {
 		$origin                = new Iri($origin);
 		$headers               = new Headers();
 		$headers['Set-Cookie'] = $header;
 
-		// Set the reference time to 2014-01-01 00:00:00
+		// Set the reference time to 2014-01-01 00:00:00.
 		$reference_time = gmmktime(0, 0, 0, 1, 1, 2014);
 
 		$parsed = Cookie::parse_from_headers($headers, $origin, $reference_time);
--- a/tests/Cookie/ParseTest.php
+++ b/tests/Cookie/ParseTest.php
@@ -510,6 +510,44 @@ final class ParseTest extends TestCase {
 	}
 
 	/**
+	 * Verify handling of Headers object with multiple `Set-Cookie` headers.
+	 *
+	 * @covers ::parse_from_headers
+	 *
+	 * @return void
+	 */
+	public function testParsingHeaderWithMultipleCookies() {
+		$origin                = new Iri('http://example.com/');
+		$headers               = new Headers();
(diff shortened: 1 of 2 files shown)


tests/Cookie/ParseTest.php
```diff
@@ -143,96 +243,145 @@ public function testParsingHeaderObject($header, $expected, $expected_attributes
...
+				'expected_attributes' => ['max-age' => 0],
+			],
+			// Invalid (non-digit character).
+			'Max-Age parsing: value: 1e6 (non-digit)' => [
+				'header'              => 'foo=bar; Max-Age=1e6',
+				'expected'            => ['expired' => false],
+				'expected_attributes' => ['max-age' => null],
+			],
+
+			/**
```
schlessera: Not an actual doc-block:
```suggestion
			/*
```

tests/Cookie/ParseTest.php
```diff
@@ -143,96 +243,145 @@ public function testParsingHeaderObject($header, $expected, $expected_attributes
...
+				'expected'            => [],
+				'expected_attributes' => ['httponly' => true],
+			],
+			'Attribute parsing: surrounding whitespace should be stripped' => [
+				'header'              => 'foo=bar; ArbitraryName   =  some-value   ',
+				'expected'            => [],
+				'expected_attributes' => ['ArbitraryName' => 'some-value'],
+			],
+
+			/**
```
schlessera: Not an actual doc-block:
```suggestion
			/*
```

tests/Cookie/ParseTest.php
```diff
@@ -143,96 +243,145 @@ public function testParsingHeaderObject($header, $expected, $expected_attributes
...
+			 * Alltogether now: multiple attributes in one cookie header.
```
schlessera: Only one 'l' for altogether.
```suggestion
			 * Altogether now: multiple attributes in one cookie header.
```

schlessera (changes requested):

schlessera (approved):

PR #765, opened 2022-07-20
Add proper usage of GHA

szepeviktor:
YAML and GHA are weird.
I use Travis CI! 🍏 

Please see comments.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,10 +15,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  #### TEST STAGE ####
+  #### TEST JOB ####
   test:
-    runs-on: ubuntu-latest
-
     strategy:
       # Keys:
       # - coverage: Whether to run the tests with code coverage.
@@ -35,49 +33,34 @@ jobs:
 
     name: "Test: PHP ${{ matrix.php }}"
     continue-on-error: ${{ matrix.php == '8.2' }}
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Set coverage variable
-        id: set_cov
-        run: |
-          if [ ${{ matrix.coverage }} == "true" ]; then
-            echo '::set-output name=COV::xdebug'
-          else
-            echo '::set-output name=COV::none'
-          fi
-
-      - name: Install PHP
+      - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
           ini-values: error_reporting=-1, display_errors=On, log_errors_max_len=0
-          coverage: ${{ steps.set_cov.outputs.COV }}
+          coverage: ${{ matrix.coverage && 'xdebug' || 'none' }}
           tools: cs2pr
 
       # At least one test needs a non-en_US locale to be available, so make sure it is.
       - name: Install locales
         run: |
           sudo apt-get update
-          sudo apt-get install locales-all
+          sudo apt-get -y install locales-all
 
       - name: Show available locales
         run: locale -a
 
       # Install dependencies and handle caching in one go.
-      # @link https://github.com/marketplace/actions/install-composer-dependencies
-      - name: Install Composer dependencies - normal
-        if: ${{ matrix.php != '8.2' }}
-        uses: "ramsey/composer-install@v2"
-
-      # For PHP "nightly", we need to install with ignore platform reqs.
       - name: Install Composer dependencies - with ignore platform
-        if: ${{ matrix.php == '8.2' }}
         uses: "ramsey/composer-install@v2"
         with:
-          composer-options: --ignore-platform-req=php
+          composer-options: ${{ matrix.php == '8.2' && '--ignore-platform-req=php' || '' }}
 
       - name: Setup problem matcher to provide annotations for PHPUnit
         run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -15,10 +15,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  #### TEST STAGE ####
+  #### TEST JOB ####
```
szepeviktor, line 18: Stages are in Travis CI :)

.github/workflows/test.yml
```diff
@@ -15,10 +15,8 @@ concurrency:
...
   test:
-    runs-on: ubuntu-latest
-
```
szepeviktor, line 21: OS version is (could be) strategy/matrix dependent: goes _after_ the matrix.

.github/workflows/test.yml
```diff
@@ -35,49 +33,34 @@ jobs:
...
-          fi
-
-      - name: Install PHP
+      - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
           ini-values: error_reporting=-1, display_errors=On, log_errors_max_len=0
-          coverage: ${{ steps.set_cov.outputs.COV }}
+          coverage: ${{ matrix.coverage && 'xdebug' || 'none' }}
```
szepeviktor, line 47: Voila! 💜

.github/workflows/test.yml
```diff
@@ -35,49 +33,34 @@ jobs:
...
-        if: ${{ matrix.php != '8.2' }}
-        uses: "ramsey/composer-install@v2"
-
-      # For PHP "nightly", we need to install with ignore platform reqs.
       - name: Install Composer dependencies - with ignore platform
-        if: ${{ matrix.php == '8.2' }}
         uses: "ramsey/composer-install@v2"
         with:
-          composer-options: --ignore-platform-req=php
+          composer-options: ${{ matrix.php == '8.2' && '--ignore-platform-req=php' || '' }}
```
szepeviktor, line 63: One step will do.

.github/workflows/test.yml
```diff
@@ -104,31 +87,21 @@ jobs:
...
-
-      - name: Show PHPUnit version
-        run: vendor/bin/phpunit --version
+      - name: Checkproxy server
+        run: |
+          mitmdump --version
+          curl -s -I http://requests-php-tests.herokuapp.com/ > /dev/null
+          ping -c1 localhost
+          curl -i http://localhost:8080
+          curl -i http://localhost:9002
```
szepeviktor, line 96: These separate steps make CI output very long but provide no additional info.

.github/workflows/test.yml
```diff
@@ -104,31 +87,21 @@ jobs:
...
 
       - name: Run the unit tests, no code coverage
-        if: ${{ matrix.coverage == false }}
+        if: ${{ ! matrix.coverage }}
```
szepeviktor, line 99: YAML has booooleans. 🎉

.github/workflows/test.yml
```diff
@@ -104,31 +87,21 @@ jobs:
...
         run: composer test
 
       - name: Run the unit tests with code coverage
-        if: ${{ matrix.coverage == true }}
-        run: vendor/bin/phpunit --coverage-clover clover.xml
+        if: ${{ matrix.coverage }}
+        run: composer exec -- phpunit --coverage-clover clover.xml
```
szepeviktor, line 104: Composer has an `exec` command. Let's use it!

.github/workflows/test.yml
szepeviktor: PHPUnit version is shown in the next step 👀

.github/workflows/test.yml
szepeviktor: I know this is ground breaking news for many ...

.github/workflows/test.yml
jrfnl: Irrelevant change. This is a left-over of when the CI was converted from Travis to GHA. The comment could even be removed.

.github/workflows/test.yml
jrfnl: Except it isn't (in the matrix), so it doesn't matter. And even if it was, the way the scripts are read by GH, you can still refer to the matrix even when the directive is _above_ the matrix definition, so this is an unnecessary change.

.github/workflows/test.yml
jrfnl: While I agree this works, I prefer readability and making it intuitive what's happening above brevity to make these kind of scripts easier to read for people less familiar with the syntax.

.github/workflows/test.yml
jrfnl: See above. While it works, it obfuscates what's happening and also makes the output of the script less readable.

With the separate steps, you can see straight away when opening a build result which of the two versions of `composer install` were run, without having to fold out the step details. This is no longer the case in your version, making the debugging of builds more cumbersome.

.github/workflows/test.yml
jrfnl: > These separate steps make CI output very long but provide no additional info.

I do not agree. If anything breaks in one of those steps, it is immediately obvious in which step the failure occurred and where to start debugging.

Debugging becomes a lot harder when multiple commands are combined in one step.

.github/workflows/test.yml
jrfnl: Yes so ? Explicit comparisons improve the readability of the script and make it more obvious what's going on.

.github/workflows/test.yml
jrfnl: Probably also groundbreaking: the `composer.json` for the project also has an explicit `coverage` script, so this could be changed to `composer coverage -- --coverage-closer clover.xml`.

Still, the overhead of running this via Composer is not worth it.

.github/workflows/test.yml
```diff
@@ -141,7 +114,7 @@ jobs:
         run: vendor/bin/stop.sh
 
       - name: Send coverage report to Codecov
-        if: ${{ success() && matrix.coverage == true }}
+        if: ${{ success() && matrix.coverage }}
```
jrfnl, line 117: See above.

jrfnl (commented):
@szepeviktor While I appreciate what you are trying to do, these are undesired changes.

Suggest: close.

.github/workflows/test.yml
szepeviktor: I know, I know... when you think about code (e.g. YAML code) it is much easier to follow line in chronological order.

.github/workflows/test.yml
szepeviktor: > for people less familiar with the syntax

I do not work for those people :(

szepeviktor:
WordPress with _people less familiar with the syntax_ is not my world.

Sorry for the PR.

.github/workflows/test.yml
jrfnl: > I do not work for those people :(

In open source you generally don't work "for" people in the first place.

However, keeping an open source project open to contributions from people with less experience than yourself is IMO always a good idea.

.github/workflows/test.yml
jrfnl: Agreed, but that would only come into play _if_ at some point the OS would be added to the matrix and this change would become relevant in that case. But not now.

PR #780, opened 2022-12-09
GH Actions: various tweaks / PHP 8.2 not allowed to fail

jrfnl:
### GH Actions: minor simplification

... of the bash `date` command in the earlier pulled cache busting.

### GH Actions: update PHP versions in workflows

PHP 8.2 has been released today 🎉 and the `setup-php` action has announced support for PHP 8.3, so adding PHP 8.3 to the matrix and no longer allowing PHP 8.2 to fail the build.

Builds against PHP 8.3 are still allowed to fail for now.

### 🆕 GH Actions: selectively use `fail-fast` with setup-php

I've seen some recent build failures (elsewhere) due to the `setup-php` action running into a rate limit and not downloading the required version of Composer. In the case of this action, without `phpdoc` being set up, the rest of the workflow would fail anyhow.

The `setup-php` action runner defaults to _showing_ these type errors in the logs, but not stopping the workflow run.

So, specifically for the update website phpdoc run, I'm adding the `fail-fast` option to `setup-php` to fail the build if the action runner ran into any errors.

Ref: https://github.com/shivammathur/setup-php#fail-fast-optional

### DIFF
--- a/.github/workflows/cs.yml
+++ b/.github/workflows/cs.yml
@@ -34,8 +34,8 @@ jobs:
       - name: Install Composer dependencies
         uses: "ramsey/composer-install@v2"
         with:
-          # Bust the cache at least once a month - output format: YYYY-MM-DD.
-          custom-cache-suffix: $(date -u -d "-0 month -$(($(date +%d)-1)) days" "+%F")
+          # Bust the cache at least once a month - output format: YYYY-MM.
+          custom-cache-suffix: $(date -u "+%Y-%m")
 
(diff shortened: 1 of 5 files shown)


jrfnl:
Note: the build failure seems unrelated to this PR, but to do with one of the endpoints we use in the tests conking out... Might be temporary, might need further investigation.

jrfnl:
I've added an extra commit to this PR with yet another GH Actions tweak.

schlessera (approved):

PR #782, opened 2022-12-12
Unblock CI

jrfnl:
CI is currently failing due to a problem with the Heroku server being used during the tests.

While that problem, of course, needs to be solved and solved ASAP, in the mean time, we do need to be able to continue work on Requests.

This PR adds a mechanism to check if the server is available and if not, skips the tests using the Heroku server.

### Tests: move the httpbin function declaration to the TestCase

... and adjust all function calls to it from within the tests.

### Tests: skip tests using `httpbin()` if the server is not available

Bit hacky, but it will work, and more importantly, it will only skip those tests which should be skipped.

### 🆕 Test bootstrap: log debug information about test server availability.

### DIFF
--- a/tests/Auth/Basic/BasicTest.php
+++ b/tests/Auth/Basic/BasicTest.php
@@ -26,7 +26,7 @@ final class BasicTest extends TestCase {
 			'auth'      => ['user', 'passwd'],
 			'transport' => $transport,
 		];
-		$request = Requests::get(httpbin('/basic-auth/user/passwd'), [], $options);
+		$request = Requests::get($this->httpbin('/basic-auth/user/passwd'), [], $options);
 
 		// Verify the request succeeded.
 		$this->assertInstanceOf(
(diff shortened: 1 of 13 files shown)


jrfnl:
This is related to Heroku no longer offering a Free service. See:
* https://blog.heroku.com/next-chapter
* https://devcenter.heroku.com/changelog-items/2461
* https://help.heroku.com/8GD2AVBW/app-down-or-not-working-after-the-termination-of-free-tier-plans-on-heroku

We'll need to look for an alternative to Heroku, preferably one to which we can automatically deploy using GH Actions from the https://github.com/RequestsPHP/test-server server.

jrfnl:
I've replaced the original second commit with a commit just and only skipping those tests which would currently fail in CI.

jrfnl:
Replaced the second commit again, but now with a better version of the original commit.

schlessera (approved):

Issue #814, opened 2023-06-26
Tests: investigate alternatives for mocking


jrfnl:
The functionality within PHPUnit itself to mock classes and set expectations on them is being deprecated left right and center, with each upcoming PHPUnit version needing more work-arounds.

While Requests doesn't use mocking much (at all), it is used in a few select places and keeping that working is getting more difficult and will probably become impossible come PHPUnit 11/12.

Typical issues:
* `setMethods()` was deprecated in PHPUnit 8.3 and removed in PHPUnit 10.
* The alternative `addMethods()` was deprecated in PHPUnit 10.1 and is expected to be removed in PHPUnit 12.
* The `getMockBuilder()` method is deprecated and is expected to be removed in PHPUnit 12.

It should be investigated if using Mockery would be an alternative (once it is available with support for PHPUnit 10).

PR #817, opened 2023-06-26
Tests: update test suite to allow for running the tests on PHPUnit 10

jrfnl:
## Pull Request Type

This is a:
- [ ] Bug fix
- [ ] New feature

## Context

Update the test suite to allow for running the tests on PHPUnit 10.

### Composer: update the PHPUnit Polyfills

A PHPUnit 10 compatible version of the PHPUnit Polyfills has been released, so let's start using it.

Ref:
* https://github.com/Yoast/PHPUnit-Polyfills/releases/tag/2.0.0

### Transport/BaseTestCase::dataSNISupport(): remove unused and unnecessary method parameter

Data providers cannot take arguments, so this `$options` parameter is useless and should never have been declared. PHPUnit 10 will flag it, so let's remove it.

### TestCase::textArrayToDataprovider(): make static

This method is used exclusively by data providers and as of PHPUnit 10, data providers should be static methods, so this method needs to be `static` too.

### TestCase::transportProvider(): make static

This method is used exclusively as a data providers and as of PHPUnit 10, data providers should be static methods, so this method needs to be `static` too.

### TestCase: split httpbin() method and make Session\ConstructorTest::dataValidUrl() method static

As of PHPUnit 10, data providers are (again) expected to be `static` methods.

One of the data providers is using the `httpbin()` method, but as the `httpbin()` method needs `$this` to handle the test skipping, the method cannot be made static.

This commit splits the `httpbin()` method into three methods:
* A `skipOnUnavailableHttpbinHost()` method which contains the test skip check.
* A `static` `getHttpbinUrl()` method which builds the URL.
* The original `httpbin()` method which now calls the two new methods to maintain the same functionality as before.

By splitting the method in this way, we prevent having to change all method calls to the `httpbin()` method throughout the test suite and can still fix the data provider method using the `httpbin()` method to not use `$this` (which allows to make it `static`).

### Tests: make dataproviders static

As of PHPUnit 10, data providers are (again) expected to be `static` methods.

This updates the test suite to respect that.

Includes removing the use of `$this` from select data providers.

### Tests: replace assertObjectHasAttribute() with `assertObjectHasProperty()`

The `assertObjectHasAttribute()` method has been deprecated in PHPUnit 9 and removed in PHPUnit 10.

PHPUnit 10.1 introduces a replacement `assertObjectHasProperty()` method, which is polyfilled by the PHPUnit Polyfills.

This switches all uses of the old method to the new method name.

### Tests: work around for MockBuilder::setMethods() removal

Please note:
* The `getMockBuilder()` method is also deprecated now, so replacing with another method call on the `MockBuilder` class will only work in the short/medium term and in a future iteration, this will have to be refactored again.
* The `addMethods()` method has been (soft) deprecated in PHPUnit 10.1 and is expected to be removed in PHPUnit 12.

Also see #814

Refs:
* sebastianbergmann/phpunit#3687#issuecomment-492537584
* sebastianbergmann/phpunit#3770
* sebastianbergmann/phpunit#3769
* sebastianbergmann/phpunit#4775

### Tests: work around for removal of expectDeprecation et al

PHPUnit will no longer stop a test on deprecations/notices/warnings.
This also means that the PHPUnit native way to expect deprecations/notices/warnings etc is no longer supported.

This implements a work-around for the limited cases in which that functionality was used in this test suite.

Refs:
* https://phpunit.de/announcements/phpunit-10.html
* sebastianbergmann/phpunit#5062

### Curl/Fsockopen: add some extra defensive coding

The `error_get_last()` function can return `null` if no error occurred, in which case, the `throw new Exception()` statement will run into two new errors:
* `Trying to access array offset on value of type null` for accessing `$error['message']`.
* ... which then leads to a `Exception::__construct(): Passing null to parameter #1 ($message) of type string is deprecated`.

This commit adds some defensive coding to handle the hypothetical situation that `error_get_last()` would return `null` when a file could not be opened.

Note: this is actually a bug in PHPUnit 10 which breaks `error_get_last()`.
We should be able to remove the extra defensive coding once the upstream bug has been fixed.

Upstream bug report: sebastianbergmann/phpunit#5428

### PHPUnit: add separate configuration for PHPUnit 10

PHPUnit 10 makes significant changes to the configuration file.

Most notably:
* In PHPUnit 9.3, the manner of specifying the code coverage configuration has changed.
* In PHPUnit 10.0, a significant number of attributes of the `<phpunit>` element were removed or renamed.
* In PHPUnit 10.1, the manner of specifying code coverage has changed again.

While there is a `--migrate-configuration` option available in PHPUnit, that doesn't get us a well enough converted configuration file, so instead use a separate `phpunit10.xml.dist` file for running the tests on PHPUnit 10 with an optimal configuration.

Includes:
* Ignoring the new file for package archives.
* Adding separate scripts to the `composer.json` file to make how to run the tests on PHPUnit 10 more obvious for contributors.
* Updating the GH Actions workflows to take the new configuration file into account when appropriate.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,3 +9,4 @@ tests/ export-ignore
 .phpcs.xml.dist export-ignore
 phpdoc.dist.xml export-ignore
 phpunit.xml.dist export-ignore
+phpunit10.xml.dist export-ignore
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -90,11 +90,17 @@ jobs:
       - name: Access localhost on port 9002
         run: curl -i http://localhost:9002
 
-      - name: Show PHPUnit version
-        run: vendor/bin/phpunit --version
+      - name: Grab PHPUnit version
+        id: phpunit_version
+        run: echo "VERSION=$(vendor/bin/phpunit --version | grep --only-matching --max-count=1 --extended-regexp '\b[0-9]+\.[0-9]+')" >> $GITHUB_OUTPUT
 
-      - name: Run the unit tests
+      - name: Run the unit tests (PHPUnit < 10)
+        if: ${{ ! startsWith( steps.phpunit_version.outputs.VERSION, '10.' ) }}
         run: composer test
+
+      - name: Run the unit tests (PHPUnit 10+)
+        if: ${{ startsWith( steps.phpunit_version.outputs.VERSION, '10.' ) }}
+        run: composer test10
 
       - name: Stop proxy server
         continue-on-error: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,16 +114,25 @@ jobs:
       - name: Access localhost on port 9002
         run: curl -i http://localhost:9002
 
-      - name: Show PHPUnit version
-        run: vendor/bin/phpunit --version
+      - name: Grab PHPUnit version
+        id: phpunit_version
+        run: echo "VERSION=$(vendor/bin/phpunit --version | grep --only-matching --max-count=1 --extended-regexp '\b[0-9]+\.[0-9]+')" >> $GITHUB_OUTPUT
 
-      - name: Run the unit tests, no code coverage
-        if: ${{ matrix.coverage == false }}
+      - name: Run the unit tests, no code coverage (PHPUnit < 10)
+        if: ${{  matrix.coverage == false && ! startsWith( steps.phpunit_version.outputs.VERSION, '10.' ) }}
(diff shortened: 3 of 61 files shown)


jrfnl:
CodeCov is drunk ;-)

schlessera (approved):

schlessera:
Re. failures:
- PHP 8.3 was expected to fail to demonstrate that deprecations are properly triggering test failures again. Will be fixed via #818 .
- Codecov failure is only an artefact, where a closing brace is improperly counted. Ignored.

PR #833, opened 2023-10-09
Add HTTP status helper class

schlessera:
This PR adds a utility class to deal with HTTP status codes and messages and then changes existing code in `Http` exceptions and the `TransportMock` to make use of this utility class.

## Pull Request Type

This is a:
- [ ] Bug fix
- [ ] New feature

## Context

PR #582 added a new `TransportRedirectMock` which was based on `TransportMock`. While doing so, it needed to duplicate the long list of status code messages already included in `TransportMock`. To make this process easier and simplify maintenance, we decided to extract knowledge about HTTP status codes and texts into a separate utility class instead, making both transport mock classes simpler in the process.

## Detailed Description

The new helper class `HttpStatus` contains two methods:

* `get_text($code)` retrieves a status text from a status code
* `is_valid_code($code)` returns a boolean on whether a given code is a valid (known) status code

> Note: This will need a follow-up PR to complete the list of HTTP status codes and texts, as the current codebase is outdated and does not contain all official HTTP status codes yet.

The helper class also contains constants that can be directly referenced, like so:
```php
$not_found_status_message = '404 ' . HttpStatus::TEXT_404;
```

The PR updates all `Http` exceptions to make use of this to de-duplicate the status text strings.

## Quality assurance

## Documentation

For new features:


src/Utility/HttpStatus.php
```diff
@@ -0,0 +1,150 @@
...
+	const TEXT_204 = 'No Content';
+	const TEXT_205 = 'Reset Content';
+	const TEXT_206 = 'Partial Content';
+	const TEXT_300 = 'Multiple Choices';
+	const TEXT_301 = 'Moved Permanently';
+	const TEXT_302 = 'Found';
+	const TEXT_303 = 'See Other';
+	const TEXT_304 = 'Not Modified';
+	const TEXT_305 = 'Use Proxy';
+	const TEXT_306 = '(Unused)';
```
jrfnl: The original text for this in the `WpOrg\Requests\Exception\Http\Status306` class was "'Switch Proxy'". Might be good to keep it that way ?

Looks like the `306` status code was originally for "Switch Proxy", but became unused with HTTP1.1.

I think if the code is still used/received, Requests should probably still handle it according to its original use ?

This probably also means that this constant needs a docblock documenting this.

src/Utility/HttpStatus.php
```diff
@@ -0,0 +1,150 @@
...
+	const TEXT_409 = 'Conflict';
+	const TEXT_410 = 'Gone';
+	const TEXT_411 = 'Length Required';
+	const TEXT_412 = 'Precondition Failed';
+	const TEXT_413 = 'Request Entity Too Large';
+	const TEXT_414 = 'Request-URI Too Long';
+	const TEXT_415 = 'Unsupported Media Type';
+	const TEXT_416 = 'Requested Range Not Satisfiable';
+	const TEXT_417 = 'Expectation Failed';
+	const TEXT_418 = 'I\'m a teapot';
```
jrfnl:
```suggestion
	const TEXT_418 = "I'm a teapot";
```

tests/Utility/HttpStatus/GetTextTest.php
```diff
@@ -0,0 +1,108 @@
...
+namespace WpOrg\Requests\Tests\Utility\HttpStatus;
+
+use WpOrg\Requests\Exception;
+use WpOrg\Requests\Exception\InvalidArgument;
+use WpOrg\Requests\Tests\TestCase;
+use WpOrg\Requests\Tests\TypeProviderHelper;
+use WpOrg\Requests\Utility\HttpStatus;
+
+/**
+ * @coversDefaultClass \WpOrg\Requests\Utility\HttpStatus
```
jrfnl:
```suggestion
 * @covers \WpOrg\Requests\Utility\HttpStatus::getText
```

As all tests in this test class are for the one method, there is no need for the individual test methods to have a `@covers` tag. A class level `@covers` tag will suffice.

(same for the other test class)

tests/Utility/HttpStatus/GetTextTest.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+class GetTextTest extends TestCase {
+	/**
```
jrfnl, line 14: Nitpick: could we have the data provider methods _after_ (directly following) the test methods ? That would improve consistency with the other (already revised) test classes.

(same for the other test class)

src/Utility/HttpStatus.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * Helper class for dealing with HTTP status codes.
+ *
+ * @package Requests\Utilities
+ */
```
jrfnl:
```suggestion
/**
 * Requests for PHP, an HTTP library.
 *
 * @copyright 2012-2023 Requests Contributors
 * @license   https://github.com/WordPress/Requests/blob/stable/LICENSE ISC
 * @link      https://github.com/WordPress/Requests
 */
```

(For consistency with #834)

src/Utility/HttpStatus.php
```diff
@@ -0,0 +1,150 @@
...
+
+namespace WpOrg\Requests\Utility;
+
+use WpOrg\Requests\Exception\InvalidArgument;
+
+/**
+ * HTTP status codes helper class.
+ *
+ * @package Requests\Utilities
+ */
```
jrfnl:
```suggestion
/**
 * Helper class for dealing with HTTP status codes.
 *
 * @package Requests\Utilities
 * @since   2.1.0
 */
```

For future maintenance, I think it might also be helpful to include a link to the (official/authoritive) source of the message texts.

src/Utility/HttpStatus.php
```diff
@@ -0,0 +1,150 @@
...
+	/**
+	 * Get the status message from a status code.
+	 *
+	 * @param int|string $code Status code.
+	 * @return string Status message.
+	 */
+	public static function get_text($code) {
+		if (self::is_valid_code($code) === false) {
+			// When the type is correct, add the value to the error message to help debugging.
+			$type = gettype($code) . ((is_int($code) || is_string($code)) ? " ($code)" : '');
```
jrfnl:
```suggestion
			$type = gettype($code) . ((is_int($code) || is_string($code)) ? " ($code)" : '');
```

This feels like we're duplicating part of the validation check again. Would `is_scalar()` suffice here ?

jrfnl (commented):
Reviewed. Left a few notes. Nothing major.

src/Utility/HttpStatus.php
```diff
@@ -0,0 +1,824 @@
...
+	 *
+	 * @var string
+	 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/102
+	 */
+	const TEXT_102 = 'Processing';
+
+	/**
+	 * HTTP response code 103 - Early Hints.
+	 *
+	 * This status code is primarily intended to be used with the Link header, letting the user agent start preloading resources while the server prepares a response or preconnects to an origin from which the page will need resources.
```
jrfnl: Nitpick:
```suggestion
	 * This status code is primarily intended to be used with the Link header, letting the user agent
	 * start preloading resources while the server prepares a response or preconnects to an origin
	 * from which the page will need resources.
```

Here and in a few of the other places where the doclines are very long/wrapping.

tests/Exception/Http/HttpTest.php
```diff
@@ -96,10 +97,100 @@ public static function dataGetClass() {
...
 		];
 	}
+
+
+	/**
+	 * Test whether all HTTP exception class can be instantiated.
+	 *
+	 * @dataProvider dataInstantiateHttpException
+	 *
+	 * @param int $code HTTP status code.
```
jrfnl:
```suggestion
	 * @param int $status_code HTTP status code.
```
Also missing `@param` tag for `$expected_exception_class`

tests/Exception/Http/HttpTest.php
```diff
@@ -96,10 +97,100 @@ public static function dataGetClass() {
...
+		$this->assertInstanceOf(Http::class, $exception_object);
+		$this->assertInstanceOf($expected_exception_class, $exception_object);
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @return array
+	 */
+	public static function dataInstantiateHttpException() {
```
jrfnl: Ha, I was just looking at this when your new commit came in. I think we can possibly make this more robust by getting rid of some of the hard-coding of the cases ?

Also using a numeric sting for the array key will not work to make these named datasets as PHP will just interpret those as integers and  PHPUnit will then not recognize them as data set names. I have a feeling we were talking about something related earlier ? 😝 

What about something like this ?
```php
		$all_status_codes    = array_keys(HttpStatus::MAP);
		$non_exception_codes = [100, 101, 102, 103, 200, 201, 202, 203, 204, 205, 206, 207, 208, 226, 300, 301, 302, 303, 307, 308];

		$data = [];
		foreach ($all_status_codes as $status) {
			if (in_array($status, $non_exception_codes, true)) {
				$data['Status ' . $status] = [$status, StatusUnknown::class];
			} else {
				$data['Status ' . $status] = [$status, Http::class . '\Status' . $status];
			}
		}

		return $data;
```

(and yes, you already have the `StatusUnknown` class imported, so no need to prefix it with `Http\`).

jrfnl: Follow up... I'm actually struggling with the validity of the test in principle as I cannot come up with a way to make the test fail....

```php
		// Invalid codes.
		$data['Status 600']            = [600, StatusUnknown::class];
		$data['Not a status - string'] = ['not a status', StatusUnknown::class];
		$data['Not a status - false']  = [false, StatusUnknown::class];
		$data['Not a status - float']  = [10.5, StatusUnknown::class];
		$data['Not a status - object'] = [new \stdClass(), StatusUnknown::class];
```

☝🏻 The above will all pass.

tests/Utility/HttpStatus/GetTextTest.php
```diff
@@ -0,0 +1,102 @@
...
+		return [
+			'negative integer' => [-1],
+			'zero integer'     => [0],
+			'too low integer'  => [42],
+			'too high integer' => [1000],
+			'negative string'  => ['-1'],
+			'zero string'      => ['0'],
+			'too low string'   => ['42'],
+			'too high string'  => ['1000'],
+		];
```
jrfnl, line 100: Should we add a test with `PHP_INT_MAX` just for fun ?

jrfnl (changes requested):
Looking good, left a few notes inline. Nothing major.

tests/Exception/Http/HttpTest.php
schlessera: The test was not meant to test input values, but rather to ensure every single class is instantiable. With exceptions, I've often seen the issue that fatal errors were only discovered in production for invalid exception classes because they were never triggered during testing. So here we're making sure that if such an exception gets triggered, it will not fatal because of a broken exception class.

jrfnl: Discussed off-GitHub: that means these tests are not actually about the `WpOrg\Requests\Exception\Http` class (which this test class is testing), but about the individual status exception classes, which means that strictly speaking, this test should be moved to its own file and get a different `@covers` tag.

jrfnl (approved):
Thanks for all your work on this @schlessera ! LGTM.

I've added two small commits still with some very minor nitpicks (mostly CS).

Do you want to clean up the branch before merge ? Or shall I merge it as is ?

jrfnl:
P.S.: build failures are due to [GH having trouble](https://www.githubstatus.com/). We'll probably best restart them once GH has recovered.

jrfnl:
Yippia-a-yeah! Ready for merge once the build finishes.

Issue #838, opened 2023-11-08
Connection: Close not set when needed - breaks downloads.


tomsommer:
#657 breaks downloads towards https://api.wordpress.org/ and many other sites when using Curl 7.29.0 (and perhaps other versions)

`Error: RuntimeException: Failed to get url 'https://api.wordpress.org/core/version-check/1.7/?locale=en_US': cURL error 28: Operation timed out after 10000 milliseconds with 807 out of -1 bytes received.`

It also causes issues with the REST API in Site Health with the error: `REST API response: (http_request_failed) cURL error 28: Operation timed out after 10005 milliseconds with XXX out of XXX bytes received”`

It also prevents WordPress plugin and core updates, basically anything that relies on the internal Curl handler in WordPress.

Adding back `Connection: Close` solves the issue. Please revert #657 ASAP and push the fix to WordPress 6.4

# If you are just looking to fix the cURL timeouts issue in WordPress 6.4 Read here:

~To fix the issue, replace `/wp-includes/Requests/src/Transport/Curl.php` with the following file: [Curl.php.zip](https://github.com/WordPress/Requests/files/13300252/Curl.php.zip) or see https://github.com/WordPress/Requests/issues/838#issuecomment-1802346004~

**UPDATE: No need to do any manual fix, updating the WordPress Core to release 6.4.1 should work just fine now and resolve the issue.**

jrfnl:
@tomsommer PR #657 fixes a bug. Reverting it recreates the bug. What makes this bug more important than the previous bug ?

I suggest someone do some research as to why the Curl version which is checked for now (7.22) is not correct/causing problems, if another version should be checked for instead (or a Curl constant) and iterate on the previous fix.

tomsommer:
@jrfnl I agree with that, I suggest the author of #657 should account for that (or the person who merged this)? Right now the change breaks previously working versions of Curl, versions which are bundled with production PHP versions (Remi PHP) and Operating Systems (RHEL7/CentOS 7), which makes this bug more important - IMHO.

jrfnl:
@tomsommer Anyone can work on this, it's open source, so contributions are welcome.

tomsommer:
For info, WP 6.4 issues mentioned in the forums in regards to timeouts, REST API problems in Site Health and issues reaching WP plugin/update servers are probably related to this issue (since curl fails on remote endpoints).
https://wordpress.org/support/forum/how-to-and-troubleshooting/
[...]

PR #657, opened 2021-12-06
Add Connection:close header only when needed

mircobabini:
## Pull Request Type

- [x] I have checked there is no other PR open for the same change.

This is a:
- [x] Bug fix
- [ ] New feature
- [ ] Code quality improvement

## Context

Fixes #656.

## Quality assurance

- [ ] This change does NOT contain a breaking change (fix or feature that would cause existing functionality to change).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added unit tests to accompany this PR.
- [ ] The (new/existing) tests cover this PR 100%.
- [x] I have (manually) tested this code to the best of my abilities.
- [x] My code follows the style guidelines of this project.

Issue #656, opened 2021-12-06
cURL incompatibility with Kaspersky (kesl)


mircobabini:
## Summary

I'm facing an issue with some servers having kesl service (Kaspersky) installed and running.

Sometimes wp_remote_* calls fails with error:
```
Scaricamento del pacchetto di installazione da https://downloads.wordpress.org/plugin/jetpack.10.1.zip...
Warning: Il download non è andato a buon fine. "cURL error 18: transfer closed with 332690 bytes remaining to read"
```

The number of bytes is completely randomic.
Since it's not consistent, I've prepared this script to be run over and over (without wp-cli cache of course):

```
while :; do  wp plugin install jetpack && wp plugin delete jetpack && rm -rf /root/.wp-cli/cache/*; done
```

After digging a lot into WpOrg\Requests\Transport\cURL implementation I've found these lines being the cause of the issue:
```php
<?php
// Force closing the connection for old versions of cURL (<7.22).
if (!isset($headers['Connection'])) {
        $headers['Connection'] = 'close';
}
```

Disabling these lines, the issue disappear. Why aren't we checking the cURL version here?
I would then suggest these changes:
1) Add a new const const CURL_7_22_0 = 0x071600
2) Check the version and add that header only if needed

## Your environment

| Environment                                                  | Answer
| ------------------------------------------------------------ | -------
| Operating system and version (e.g. Ubuntu 20.04, CentOS, tried both)
| PHP version                                                  | 7.4 (tried others as well)
| Requests version                                             | provided with WordPress latest
| Autoloader used                                              | WordPress
| Link to your project                                         | into a vpn, not accessible through the internet

## Tested against `develop` branch?
- [ ] I have verified the issue still exists in the `develop` branch of Requests.

mircobabini:
As reported here: https://core.trac.wordpress.org/ticket/54588

jrfnl:
FYI: A patch for this issue was included in the Requests 2.0.8 release and reverted in Requests 2.0.9 due to it causing problems with Curl 7.29.0 (and possibly others). See #838

Re-opening this issue as it is now unsolved again. This will need further investigation in conjunction with #838.

PR #839, opened 2023-11-08
Rollback changes from PR #657

laszlof:
This change rolls back a bug fix from PR #657. This PR contained no test's, and no real documentation what the issue was it was trying to solve. As a result, this change has broken updates (amongst other things) on at least anyone using curl 7.29.

See Issue #838 for more details.

## Pull Request Type

- [X] I have checked there is no other PR open for the same change.

This is a:
- [X] Bug fix
- [ ] New feature

## Context
Roll back a bug fix from PR #657. 

## Quality assurance

- [X] This change does NOT contain a breaking change (fix or feature that would cause existing functionality to change).
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have (manually) tested this code to the best of my abilities.
- [X] My code follows the style guidelines of this project.

## Documentation

For new features:

### DIFF
--- a/src/Transport/Curl.php
+++ b/src/Transport/Curl.php
@@ -27,7 +27,6 @@ use WpOrg\Requests\Utility\InputValidator;
 final class Curl implements Transport {
 	const CURL_7_10_5 = 0x070A05;
 	const CURL_7_16_2 = 0x071002;
-	const CURL_7_22_0 = 0x071600;
 
 	/**
 	 * Raw HTTP data
@@ -383,7 +382,7 @@ final class Curl implements Transport {
 		$options['hooks']->dispatch('curl.before_request', [&$this->handle]);
 
 		// Force closing the connection for old versions of cURL (<7.22).
-		if ($this->version < self::CURL_7_22_0 && !isset($headers['Connection'])) {
+		if (!isset($headers['Connection'])) {
 			$headers['Connection'] = 'close';
 		}
 


schlessera (approved):

jrfnl:
For the record: the underlying issue #656 contains information on what #657 solves.

src/Transport/Curl.php
```diff
@@ -383,7 +382,7 @@ private function setup_handle($url, $headers, $data, $options) {
 		$options['hooks']->dispatch('curl.before_request', [&$this->handle]);
 
 		// Force closing the connection for old versions of cURL (<7.22).
```
sybrew, line 384: This comment is no longer valid.

schlessera: The comment was already present before we made the problematic change and was the reason why we used the specific version constraint we ended up with. Before we fully understand the context and where this was originally coming for, I'll prefer to keep in in there.

tomsommer: The version number in the comment could be a cause for confusion going forward, just a heads up.

schlessera: Yes, indeed, and I think that's what has already happened in this case.
But I'm still assuming there is a reason why this was put in there initially. I'd like to understand this better before removing it.

schlessera: For future reference, this is the PR that introduced the comment: https://github.com/WordPress/Requests/pull/211

PR #847, opened 2023-12-02
Allow integers as cookie names

nosilver4u:
Fixes https://github.com/WordPress/Requests/issues/845 so that an exception isn't thrown for legitimate numeric cookie names.

Changes for both the constructor and parse() methods:
- Modifies the check(s) for $name, to allow both string and integer cookie names.
- Changes the exception thrown to indicate a string OR int is permitted.
- Updates doc blocks to allow cookie names to be integers.
- Modifies unit tests for $name parameter to allow strings and integers, and to validate the updated exception messages.

I have kept Cookie->name as a string, and cast $name to a string within the constructor, just in case some (external) code would be doing a strict comparison against the name, and expecting a string value. If that's not desirable, I can revert that change.

## Pull Request Type

- [x] I have checked there is no other PR open for the same change.

This is a:
- [x] Bug fix
- [ ] New feature
- [ ] Documentation improvement
- [ ] Code quality improvement

## Context

Currently, multiple plugins, and possibly even WP core are throwing errors if someone has a cookie with a numeric name. Would love to see this issue go away, so that folks don't have to resort to deleting cookies in the dev tools of their browser.

## Detailed Description
This issue typically comes up in plugins that are executing 'async' requests, such that the existing cookies ($_COOKIES) are copied into the 'cookies' parameter for a GET/POST request. One could sanitize the numeric key names at that point, but PHP will cast the indices back to integers, and then an exception will still be thrown.

## Quality assurance

- [x] This change does NOT contain a breaking change (fix or feature that would cause existing functionality to change).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added unit tests to accompany this PR.
- [x] The (new/existing) tests cover this PR 100%.
- [x] I have (manually) tested this code to the best of my abilities.
- [x] My code follows the style guidelines of this project.

Issue #845, opened 2023-11-30
Cookies with numeric/integer names cause exception.


nosilver4u:
## Summary

As reported in https://core.trac.wordpress.org/ticket/58566 when a cookie with a numeric name is used in a request, an exception is thrown. Though a cookie with a numeric name seems odd to me, it is permitted by the spec, and happens "in the wild".

### Given the following code sample

```php
$result = WpOrg\Requests\Requests::post(
        'https://example.com/submit/',
        array(),
        array(  
                'test_verify' => 'abc123',
        ),      
        array(  
                'cookies' => array(
                        'stuff' => 'good one!',
                        5       => 'stuff',
                )       
        )       
);
```

### I'd expect the following behaviour
The request to succeed.

### Instead this happened
```
PHP Fatal error:  Uncaught WpOrg\Requests\Exception\InvalidArgument: WpOrg\Requests\Cookie::parse(): Argument #2 ($name) must be of type string, integer given in rmccue/requests/src/Exception/InvalidArgument.php:29
Stack trace:
#0 rmccue/requests/src/Cookie.php(422): WpOrg\Requests\Exception\InvalidArgument::create(2, '$name', 'string', 'integer')
#1 rmccue/requests/src/Cookie/Jar.php(61): WpOrg\Requests\Cookie::parse('stuff', 5)
#2 rmccue/requests/src/Cookie/Jar.php(156): WpOrg\Requests\Cookie\Jar->normalize_cookie('stuff', 5)
#3 rmccue/requests/src/Hooks.php(93): WpOrg\Requests\Cookie\Jar->before_request(Object(WpOrg\Requests\Iri), Array, Array, 'POST', Array)
#4 rmccue/requests/src/Requests.php(455): WpOrg\Requests\Hooks->dispatch('requests.before...', Array)
#5 rmccue/requests/src/Requests.php(342): WpOrg\Requests\Requests::request('https://example...', Array, Array, 'POST', Array)
#6 functions.php(82): WpOrg\Requests\Requests::post('https://example...', Array, Array, Array)
...
#13 {main}
  thrown in /Users/sbishop/Downloads/ewwwio-silo/vendor/rmccue/requests/src/Exception/InvalidArgument.php on line 29
```
### Additional context
To my knowledge, this affects several plugins making async requests in WordPress, including EWWW Image Optimizer and WooCommerce, and also possibly WP core (site health check).

## Your environment
| Environment                                                  | Answer
| ------------------------------------------------------------ | -------
[...]

PR #849, opened 2023-12-11
New `InputValidator::is_valid_rfc2616_token()` method

jrfnl:
## Pull Request Type

This is a:
- [ ] Bug fix
- [x] New feature

## Context

The new method is intended to be used as a helper to allow for making solving #845 / #847 more straight-forward.

## Detailed Description

This new method validates that an arbitrary input parameter can be considered valid for use as a "token" as per the RFC 2616 specification.

Includes tests.

Ref: https://datatracker.ietf.org/doc/html/rfc2616#section-2.2

### DIFF
--- a/src/Utility/InputValidator.php
+++ b/src/Utility/InputValidator.php
@@ -114,7 +114,7 @@ final class InputValidator {
 	 * Verify that a received input parameter is a valid "token name" according to the
 	 * specification in RFC 2616 (HTTP/1.1).
 	 *
-	 * The short version is: 1 or more ascii characters, CTRL chars and separators not allowed.
+	 * The short version is: 1 or more ASCII characters, CTRL chars and separators not allowed.
 	 * For the long version, see the specs in the RFC.
 	 *
 	 * @link https://datatracker.ietf.org/doc/html/rfc2616#section-2.2
--- a/src/Utility/InputValidator.php
+++ b/src/Utility/InputValidator.php
@@ -109,4 +109,27 @@ final class InputValidator {
 
 		return false;
 	}
+
+	/**
+	 * Verify that a received input parameter is a valid "token name" according to the
+	 * specification in RFC 2616 (HTTP/1.1).
+	 *
+	 * The short version is: 1 or more ascii characters, CTRL chars and separators not allowed.
+	 * For the long version, see the specs in the RFC.
+	 *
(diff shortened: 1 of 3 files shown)


src/Utility/InputValidator.php
```diff
@@ -109,4 +109,27 @@ public static function is_curl_handle($input) {
 
 		return false;
 	}
+
+	/**
+	 * Verify that a received input parameter is a valid "token name" according to the
+	 * specification in RFC 2616 (HTTP/1.1).
+	 *
+	 * The short version is: 1 or more ascii characters, CTRL chars and separators not allowed.
```
schlessera: I'd capitalize the ASCII acronym, it is not an actual word.
```suggestion
	 * The short version is: 1 or more ASCII characters, CTRL chars and separators not allowed.
```

src/Utility/InputValidator.php
```diff
@@ -109,4 +109,27 @@ public static function is_curl_handle($input) {
...
+	 * @param mixed $input Input parameter to verify.
+	 *
+	 * @return bool
+	 */
+	public static function is_valid_rfc2616_token($input) {
+		if (!is_int($input) && !is_string($input)) {
+			return false;
+		}
+
+		return (preg_match('@^[0-9A-Za-z!#$%&\'*+.^_`|~-]+$@', $input) === 1);
```
schlessera: These outer parentheses are superfluous here. Was that a conscious decision for clarity?
For me, this is tripping me up here in terms of readability as I initially tried to figure out what I was missing...

schlessera (changes requested):

schlessera:
@jrfnl I've made the changes we've discussed in the call. Do you want to take a final look?

schlessera (approved):

src/Utility/InputValidator.php
jrfnl: As discussed off-GH: Yes, conscious in the sense of force of habit, as I've seen issues with operator precedence too often.

Not an issue here, so I'm fine with removing them.

jrfnl:
@schlessera Thanks for working on this together. all looks good to me.

jrfnl:
@schlessera Shall I clean up the branch before merge ? Or will you squash merge ?

PR #865, opened 2024-03-15
[PHP 8.4] Fixes for implicit nullability deprecation

Ayesh:
## Pull Request Type

This is a:
- [x] Bug fix
- [ ] New feature

## Context

Fixes all issues that emit deprecation notices on PHP 8.4 for implicit nullable parameter type declarations.

See:
 - [RFC](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)
 - [PHP 8.4: Implicitly nullable parameter declarations deprecated](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)

## Quality assurance

## Documentation

For new features:

### DIFF
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -492,7 +492,22 @@ class Cookie {
 	 * @param int|null                         $time    Reference time for expiration calculation
 	 * @return array
 	 */
-	public static function parse_from_headers(Headers $headers, Iri $origin = null, $time = null) {
+	public static function parse_from_headers(Headers $headers, $origin, $time = null) {
+		// Mimic a \WpOrg\Requests\Iri|null type-check for $origin.
+		// Change to a type declaration when PHP >= 7.1 is the base requirement.
+		if ($origin !== null && !$origin instanceof \WpOrg\Requests\Iri) {
+			if (\PHP_VERSION_ID >= 70000) {
+				throw new \TypeError(
+					'WpOrg\Requests\Cookie::parse_from_headers(): Argument #2 ($origin) must be of type WpOrg\Requests\Iri, ' . gettype($origin) . ' given'
+				);
+			}
+			trigger_error(
+				'Argument 2 passed to WpOrg\Requests\Cookie::parse_from_headers() must be an instance of WpOrg\Requests\Iri, ' . gettype($origin) . ' given',
+				E_USER_ERROR
+			);
+			exit();
+		}
+
 		$cookie_headers = $headers->getValues('Set-Cookie');
 		if (empty($cookie_headers)) {
 			return [];
--- a/src/Cookie.php
+++ b/src/Cookie.php
@@ -491,26 +491,17 @@ class Cookie {
 	 * @param \WpOrg\Requests\Iri|null         $origin  URI for comparing cookie origins
 	 * @param int|null                         $time    Reference time for expiration calculation
 	 * @return array
+	 *
+	 * @throws \WpOrg\Requests\Exception\InvalidArgument When the passed $origin argument is not null or an instance of the Iri class.
 	 */
-	public static function parse_from_headers(Headers $headers, $origin, $time = null) {
-		// Mimic a \WpOrg\Requests\Iri|null type-check for $origin.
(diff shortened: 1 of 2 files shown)


Art4:
Will this require to drop support for PHP 5.6 and 7.0?

Ayesh:
Ah, I thought this library also requires the same PHP version as WordPress. 
But yes, you are right if this is merged, it will break on PHP < 7.1.

jrfnl:
@Ayesh Thanks for the PR.

@Ayesh @Art4 WordPress still has a minimum of PHP 7.0, this package of PHP 5.6, so this change can not go in.

I'd already been looking at this one and have another patch prepared, but hadn't pulled it yet as I wanted to discuss it with @schlessera first.

We will basically need to make a choice between:
1. Dropping support for PHP < 7.1 and adding the nullable operator. Unlikely that would be acceptable until WP would drop support for PHP < 7.1.
2. Making the parameter required `Iri $origin`. This would be a BC break which impacts all _calls_ to the method.
3. Removing the type `$origin = null` and validating the parameter (if passed) to be `instanceof Iri` within the function itself. Still a BC break as the class is not `final` and the method is not `final` either, so it could be overloaded from child classes, which would now get a signature mismatch, but a BC-break for which the impact would be smaller as the method is not overloaded that often (I expect).

I'm leaning towards option 3 (which is the patch I prepared locally).

Ayesh:
Thank you @jrfnl - you are absolutely right making the parameter type nullable was not going for to work on PHP < 7.1.
I tried again in the 3rd approach you mentioned. I think that's the best thing we can do apart from bumping minimum requirement to PHP 7.1+.

I tried to recreate error handling, it may not be perfect, but it tries to mimic PHP 5.6 behavior as well.

src/Cookie.php
```diff
@@ -492,7 +492,22 @@ public static function parse($cookie_header, $name = '', $reference_time = null)
...
+				throw new \TypeError(
+					'WpOrg\Requests\Cookie::parse_from_headers(): Argument #2 ($origin) must be of type WpOrg\Requests\Iri, ' . gettype($origin) . ' given'
+				);
+			}
+			trigger_error(
+				'Argument 2 passed to WpOrg\Requests\Cookie::parse_from_headers() must be an instance of WpOrg\Requests\Iri, ' . gettype($origin) . ' given',
+				E_USER_ERROR
+			);
+			exit();
+		}
```
jrfnl: I think this can be simplified by using the `WpOrg\Requests\Exception\InvalidArgument` exception, which we're also using everywhere else we're doing input validation in-function in Requests:

```php
		if (!empty($origin) && !($origin instanceof Iri)) {
			throw InvalidArgument::create(2, '$origin', Iri::class, gettype($origin));
		}
```

Also take note of the parentheses around the `instanceof`. These are necessary as otherwise the precedence order will cause this code to be read as `(!$origin) instanceof Iri`, which is effectively `true|false instanceof Iri`, which would always be `false`.

The function will now also need to get a `@throws` tag.

As a last point, I wonder if this check should be moved to below the `$cookie_headers` check ? as that can function just fine without the `$origin` being validated.

jrfnl (commented):
@Ayesh Thank you for making those updates.

I've left some feedback in line for you to consider.

src/Cookie.php
jrfnl: Oh and before I forget: this change will also need a test to cover it. Basically a test passing a non-object and one passing a non-Iri object to the method to make sure those get rejected correctly. The `WpOrg\Requests\Tests\TypeProviderHelper` can be used to set up the data provider for this test to make it comprehensive.

jrfnl:
@Ayesh FYI: Alain and me have discussed this ticket and I've updated the PR to bring back the `null` default value, add the `@throws` tag, simplify the condition and add the test.

Ayesh:
Looking great, thank you!

schlessera (approved):

jrfnl:
Thanks @Ayesh for working on this. The patch is included in today's 2.0.11 release.

PR #874, opened 2024-06-03
Add testing proxy temporary files to .gitignore

schlessera:
The testing proxy we use for testing actual HTTPS remote requests (via Python's `mtmproxy`) creates temporary files under the `tests/utils/proxy` folder:

* a cache folder named `__pycache__`
* PID files to store references to the proxy server processes

This PR adds these to the `.gitignore` files so they don't get accidentally committed to GitHub.


jrfnl (approved):
:shipit:

PR #878, opened 2024-07-08
Mark parameters containing user credentials as sensitive

jrfnl:
PHP 8.2 introduced the `SensitiveParameter` attribute.

The effect of the attribute is that the value of the parameter is no longer directly shown in stack traces; instead, starting with PHP 8.2, the parameter will be presented as a `SensitiveParameterValue` object.

As the attribute only applies to parameters, it (unfortunately) has no effect on serialization of the object. See: https://3v4l.org/StoQO Might be an idea to start a discussion about an `SensitiveProperty` attribute on the PHP Internals mailing list, but that's outside the scope of this PR.

For now, this PR marks the `$args` parameter for the `Auth\Basic` class constructor and the `Proxy\Http` constructor as sensitive as both of these are supposed to contain user credentials (user name, password) for accessing a protected URL.

Includes updating the example code for custom authentication to also use the attribute.

### Open question

The `$options` array passed to a large range of Requests methods can [also contain credentials](https://github.com/WordPress/Requests/blob/ebb9f65855c860bc33005b3d8bccf6444e598fba/src/Requests.php#L395-L399). Should this parameter also be marked as sensitive in all appropriate places ?

Refs:
* https://www.php.net/manual/en/class.sensitiveparameter.php
* https://wiki.php.net/rfc/redact_parameters_in_back_traces

### DIFF
--- a/docs/authentication-custom.md
+++ b/docs/authentication-custom.md
@@ -16,7 +16,10 @@ services that do this; perhaps this is a market waiting to be tapped?)
 class MySoftware_Auth_Hotdog implements WpOrg\Requests\Auth {
     protected $password;
 
-    public function __construct($password) {
+    public function __construct(
+        #[\SensitiveParameter]
+        $password
+    ) {
         $this->password = $password;
     }
 
--- a/src/Auth/Basic.php
+++ b/src/Auth/Basic.php
@@ -9,6 +9,7 @@
 
 namespace WpOrg\Requests\Auth;
 
+use SensitiveParameter;
 use WpOrg\Requests\Auth;
 use WpOrg\Requests\Exception\ArgumentCount;
 use WpOrg\Requests\Exception\InvalidArgument;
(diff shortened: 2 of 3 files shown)


schlessera:
> The $options array passed to a large range of Requests methods can [also contain credentials](https://github.com/WordPress/Requests/blob/ebb9f65855c860bc33005b3d8bccf6444e598fba/src/Requests.php#L395-L399). Should this parameter also be marked as sensitive in all appropriate places ?

Wrapping the entire `$options` array as a sensitive parameter would hinder a lot of the debugging efforts, so I think that might be too generic to block.

How about creating a polyfill for `SensitiveParameterValue` on PHP < 8.2 instead and wrapping only `$options>auth` and `$options>proxy` in that class?

See https://www.php.net/manual/en/class.sensitiveparametervalue.php

jrfnl:
Let's think this over some more.

PR #881, opened 2024-07-08
CS/QA: various minor fixes

jrfnl:
### Response\Headers\FlattenTest: minor improvement to data provider

... for readability.

### Docs: favour generic type hint syntax over array type hint syntax

... for higher specificity and improved compatibility with tools like PHPStan.

### Docs: various minor tweaks

* Punctuation and spelling fixes.

### Transport/BaseTestCase: make child class expectation explicit

The `BaseTestCase` class expects a child class to declare a `$transport` property. This makes that expectation explicit.

### Response/Headers/GetIteratorTest: minor fix

The `WpOrg\Requests\Exception::__construct()` method expects a minimum of 2 (required) parameters.

While this `default` case is basically impossible to be hit, the exception creation should still be correct.

### Cookie/ParseTest: fix an incorrect condition

As the `$expected_attributes` and `$expected_flags` variables are optional parameters, these will always be set. However, they may not be set to an array, which is what this condition should protect against.

### DIFF
--- a/src/Auth/Basic.php
+++ b/src/Auth/Basic.php
@@ -76,9 +76,9 @@ class Basic implements Auth {
 	}
 
 	/**
-	 * Set cURL parameters before the data is sent
+	 * Set cURL parameters before the data is sent.
 	 *
-	 * @param resource|\CurlHandle $handle cURL handle
+	 * @param resource|\CurlHandle $handle The cURL handle.
 	 */
 	public function curl_before_send(&$handle) {
 		curl_setopt($handle, CURLOPT_HTTPAUTH, CURLAUTH_BASIC);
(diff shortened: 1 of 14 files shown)


src/Auth/Basic.php
```diff
@@ -76,19 +76,19 @@ public function register(Hooks $hooks) {
 	}
 
 	/**
-	 * Set cURL parameters before the data is sent
+	 * Set cURL parameters before the data is sent.
 	 *
-	 * @param resource|\CurlHandle $handle cURL handle
```
schlessera, line 81: AFAIK, the name of the project is `cURL` => https://curl.se/docs/faq.html#What_is_cURL

Lowercase `curl` would be the shell binary, but the handle here is actually for the integrated `libcurl` library, not for the shell binary.

The class name `CurlHandle` comes from using PascalCase for PHP class names while ignoring brand names and trademarks (like `Mysql` vs `MySQL`).

src/Transport/Curl.php
```diff
@@ -20,7 +20,7 @@
 use WpOrg\Requests\Utility\InputValidator;
 
 /**
- * cURL HTTP transport
+ * Curl HTTP transport.
```
schlessera: See discussion above about `cURL` project name.

schlessera (changes requested):

schlessera (approved):

PR #916, opened 2024-12-11
Streamline local test environment setup

### LINKED ISSUE #819: Tests: Streamline running of tests against all PHPUnit versions
The PR #817 introduced a new configuration file for PHPUnit 10, separate from the previous configuration file. In order to run tests against PHPUnit 10, you now have to run `composer test10` instead of only `composer test`.

It would be preferable to have only a singular entry point of `composer test` that is smart enough to run PHPUnit against the right configuration file.

While looking into this, we've found that PHPUnit has [a very convenient flag `--atleast-version <version>`](https://docs.phpunit.de/en/10.0/textui.html#miscellaneous) which could be used to run the tests conditionally. So, the idea is to have something like this (simplified):

```
    "test": [
      "phpunit --atleast-version 10 && phpunit -c phpunit10.xml.dist --no-coverage",
      "phpunit --atleast-version 10 || phpunit --no-coverage"
    ],
```

That should generally work.

However, we have yet to find a way to combine this with the use of `@php` to reuse the exact PHP process that Composer is running under. According to the [Composer documentation](https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts), you cannot combine multiple commands like this:

> One limitation of this is that you can not call multiple commands in a row like @php install && @php foo. You must split them up in a JSON array of commands.

This is why PR #817 has separate Composer scripts at this point: `composer test` & `composer test10`.

This goes against Composer conventions and requires the user to know details about the test flow and dependencies in order to run the tests.

schlessera:
Refactored the test environment setup to make it more robust and user-friendly. This consolidates multiple test commands and improves process management for local development.

Fixes #819 

## What Changed

* Added shell scripts to manage test environment lifecycle
* Consolidated PHPUnit commands to auto-detect version and config
* Improved process management with proper PID tracking
* Added mitmproxy version verification (≥11.0.2 required)
* Adapted `CONTRIBUTING.md` to explain the new setup

The main improvement is the simplified test workflow. Instead of juggling different commands for PHPUnit versions and remembering to start/stop test servers, everything is now handled through clear, single-purpose commands.

### New Scripts

* `scripts/start-test-environment.sh` - Boots test and proxy servers - should be `source`d to set environment vairables
* `scripts/stop-test-environment.sh` - Clean shutdown of all processes
* `scripts/run-phpunit.sh` - Version-agnostic PHPUnit runner

## Usage

Basic testing is now simpler:

```bash
composer test # Run tests (auto-detects PHPUnit version)
composer coverage # Same, but with coverage
```

For development with local test servers:

```bash
# Direct usage (sets env vars in your shell)
source scripts/start-test-environment.sh
scripts/run-phpunit.sh
scripts/stop-test-environment.sh

# Or through composer (automated start/stop)
composer test:withserver
composer coverage:withserver
```


jrfnl (commented):
Looking good.

Will still need a note in CONTRIBUTING that this will not work on Windows for now.

As discussed, not merging yet until a shellcheck check has been added in CI to safeguard the scripts.

jrfnl:
Future scope: add a Docker config file to get up & running with even less effort (and which would work on Windows too).

jrfnl:
Just noticed, just running the script on my machine created some `.test-pids/*.pid` files. You may want to add those to the `.gitignore` file ?

schlessera:
I changed the folder where PIDs are stored and added that location to `.gitignore`: `tests/utils/pids`

jrfnl:
Just checking - this PR relates to #819 (and closes it ?)

.github/CONTRIBUTING.md
```diff
@@ -109,28 +109,65 @@ Code coverage is monitored for every PR and for the code base as a whole using [
...
+
+Note: The environment scripts must be sourced (using `source` or `.`) to properly set environment variables in your shell.
+
+#### 3. Individual Component Control
+
+For debugging or development, you might want to manage individual components:
+
 ```bash
-# Start the test server
+# Just the test server
```
jrfnl:
```suggestion
# Start only the test server
```

.github/CONTRIBUTING.md
```diff
@@ -109,28 +109,65 @@ Code coverage is monitored for every PR and for the code base as a whole using [
...
 PORT=8080 vendor/bin/start.sh
-export "REQUESTS_TEST_HOST_HTTP=localhost:8080"
+export REQUESTS_TEST_HOST_HTTP=localhost:8080
 
-# Start the proxy server
+# Just the proxy servers
```
jrfnl:
```suggestion
# Start only the proxy servers
```

.github/workflows/quicktest.yml
```diff
@@ -94,14 +94,9 @@ jobs:
         id: phpunit_version
         run: echo "VERSION=$(vendor/bin/phpunit --version | grep --only-matching --max-count=1 --extended-regexp '\b[0-9]+\.[0-9]+')" >> $GITHUB_OUTPUT
```
jrfnl: Looks like this step can now be removed ?

.github/CONTRIBUTING.md
```diff
@@ -109,28 +109,65 @@ Code coverage is monitored for every PR and for the code base as a whole using [
...
-export "REQUESTS_HTTP_PROXY_AUTH_USER=test"
-export "REQUESTS_HTTP_PROXY_AUTH_PASS=pass"
+export REQUESTS_HTTP_PROXY=localhost:9002
+export REQUESTS_HTTP_PROXY_AUTH=localhost:9003
+export REQUESTS_HTTP_PROXY_AUTH_USER=test
+export REQUESTS_HTTP_PROXY_AUTH_PASS=pass
+```
 
-# Run the tests
-composer test
```
jrfnl, line 126: Why is this being removed here ?

.github/workflows/test.yml
```diff
@@ -118,22 +118,14 @@ jobs:
         id: phpunit_version
         run: echo "VERSION=$(vendor/bin/phpunit --version | grep --only-matching --max-count=1 --extended-regexp '\b[0-9]+\.[0-9]+')" >> $GITHUB_OUTPUT
```
jrfnl: Dito. This step can probably be removed now.

composer.json
```diff
@@ -87,25 +87,29 @@
...
+    "testserver:start": [
+      "./scripts/start-test-environment.sh"
+    ],
+    "testserver:stop": [
+      "./scripts/stop-test-environment.sh"
     ],
-    "coverage10": [
-      "@php ./vendor/phpunit/phpunit/phpunit -c phpunit10.xml.dist"
+    "test:withserver": [
+      "./scripts/run-tests-with-server.sh"
```
jrfnl: Out of curiosity: why is this a separate shell script instead of combining existing scripts in the composer config ?

```yml
    "test:withserver": [
        "@testserver:start",
        "@test",
        "@testserver:stop"
    ],
```

Is it because of exit code handling ?

Also: why is there only a `test:withserver`, no `coverage:withserver` ?

scripts/run-phpunit.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+# Detect PHPUnit version
+PHPUNIT_VERSION=$("${PROJECT_ROOT}/vendor/bin/phpunit" --version | grep -oE '[0-9]+\.[0-9]+' | head -n1)
+
+# Run the tests with the appropriate config
+if printf '%s\n%s\n' "10.0" "$PHPUNIT_VERSION" | sort -V -C 2>/dev/null; then
+    "${PROJECT_ROOT}/vendor/bin/phpunit" -c phpunit10.xml.dist "$@"
+else
+    "${PROJECT_ROOT}/vendor/bin/phpunit" -c phpunit.xml.dist "$@"
```
jrfnl: This is a regression (also line 15). The commands used here will use the system default PHP version instead of the Composer PHP version, which was used previously (and deliberately set up to be used).

scripts/run-phpunit.sh
```diff
@@ -0,0 +1,22 @@
...
+# Try to determine script location
+if [ -n "${BASH_SOURCE:-}" ]; then
+    # Bash-specific path resolution
+    SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+    PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
+else
+    # POSIX fallback - assume we're in the repository root
+    PROJECT_ROOT="$(pwd)"
+    SCRIPT_DIR="$PROJECT_ROOT/scripts"
+fi
```
jrfnl: Out of curiousity: why is this needed ? AFAIK Composer scripts only work from the directory containing the `composer.json` file, so in that case, we'd always be in the project root.

I imagine it's because people can call these scripts directly ?

scripts/run-tests-with-server.sh
```diff
@@ -0,0 +1,26 @@
...
+if ! . "${SCRIPT_DIR}/start-test-environment.sh"; then
+    exit 1
+fi
+
+# Run the tests
+"${SCRIPT_DIR}/run-phpunit.sh" --no-coverage
+TEST_EXIT_CODE=$?
+
+# Stop the test environment
+. "${SCRIPT_DIR}/stop-test-environment.sh"
```
jrfnl: Shouldn't there be an error as well if the test environment couldn't be stopped cleanly ?

scripts/start-test-environment.sh
```diff
@@ -0,0 +1,120 @@
...
+    if [ -n "${BASH_VERSION:-}" ]; then
+        exit 1
+    fi
+fi
+
+PID_DIR="${PROJECT_ROOT}/tests/utils/pids"
+
+# Check if mitmproxy is installed
+if ! command -v mitmdump >/dev/null 2>&1; then
+    echo "Error: mitmproxy is not installed. Please install it with: pip3 install mitmproxy"
```
jrfnl, line 55: Should there be a check first to see if Python/Pip3 is installed ?

scripts/start-test-environment.sh
jrfnl, line 1: Just wondering: if a user already has a server running which serves `localhost`, will this script cause problems with that?

.github/CONTRIBUTING.md
```diff
@@ -109,28 +109,65 @@ Code coverage is monitored for every PR and for the code base as a whole using [
 
 ### Running the Tests
```
jrfnl, line 111: Might be good to update the "Prerequisites for running the tests" section to mention the minimum versions, i.e. Python 3.12 and mitmproxy 11.0.2.

jrfnl (changes requested):
@schlessera I've had a good look at this PR, with the caveat of limited knowledge of bash scripting... 

Left a number of questions in-line.

Additionally:
* The scripts re-introduce a problem we previously solved: the system default PHP version being used instead of the Composer PHP version being used. This is a regression, which I'm not happy with.
* `/scripts/` directory needs to be added to `.gitattributes`
* `composer.json`: script descriptions should mention that this doesn't work on Windows, only on *nix
* `CONTRIBUTING`: should mention that this doesn't work on Windows, only on *nix.
* `CONTRIBUTING`: "The environment scripts must be sourced"...  what does that mean ?
    Note: the `source` command is not available on Windows.

Other remarks:
* The minimum versions for Python and mitmproxy will now need to be maintained in quite a few places (`CONTRIBUTING`, `quicktest.yml`, `test.yml`, the new scripts). This feels pretty error prone...
* Just as a test, I've ran the scripts on Windows:
    - `run-tests-with-server.sh` doesn't appears to do anything (other than flash open a new shell window & close it) and exits without giving the user information about why it's not working.
    - `start-test-environment.sh` doesn't appears to do anything (other than flash open a new shell window & close it) and exits without giving the user information about why it's not working.
    - `stop-test-environment.sh` doesn't appears to do anything (other than flash open a new shell window & close it) and exits without giving the user information about why it's not working.
    - `run-phpunit.sh` runs, but opens a new shell window and closes it immediately once the test run has finished, meaning you can't read the results.
        The new window is also really annoying for other reasons, like usability issues, no scrolling, window size being tiny etc

scripts/run-phpunit.sh
schlessera: This was meant to allow the use of `composer --working-dir=...` to work properly. However, with all the iterations I've done, this seems to have broken anyway, as I can't make it work correctly right now. I'll just remove this bit for now and we can always readd later if we have an actual issue that needs solving.

composer.json
schlessera: Because of the way Composer works, when you combine the existing scripts, the environment variables defined in one are not forwarded to the next one. So we have no way of letting one scripts creating servers and then letting the next script know where to find them (unless we'd create file-based references somewhere).

By putting this into its own script, we can control the environment variables across the different steps.

However, there is an alternative where we use `@putenv` to manually hardcode the environment variables into the script. Given that the defaults are hardcoded anyway for these steps, this would have the advantage of being less platform-specific.

Let me rewrite this and see if that is better...

scripts/run-tests-with-server.sh
schlessera: I'll have to think how best to do that, let me figure something out...

scripts/start-test-environment.sh
schlessera: I'd consider that out of scope. If you can get mitmproxy to run without having pip around or python properly set up, we should'nt break. We only need mitmproxy to work _somehow_.

scripts/start-test-environment.sh
schlessera: Only if there is a conflict with the ports. So if you have a WP site running on `localhost:80`, for example (default web server), that would be fine.
We could make the ports configurable through the simplified scripts, but they are already configurable through the manual route, so I'd think that is flexible enough.

composer.json
schlessera: Changed this to use `@putenv`.

scripts/run-phpunit.sh
schlessera: Fixed.

.github/CONTRIBUTING.md
schlessera: Because this section is now focused on how you control (i.e. start and stop) the individual test environment components. What you do in-between is probably much more complex than just running the tests if you need that much control.

scripts/start-test-environment.sh
```diff
@@ -0,0 +1,125 @@
...
+echo $! > "${PID_DIR}/proxy-server.pid"
+
+PORT=9003 AUTH="test:pass" "${PROJECT_ROOT}/tests/utils/proxy/start.sh"
+echo $! > "${PID_DIR}/proxy-auth-server.pid"
+
+# Set environment variables
+REQUESTS_HTTP_PROXY="localhost:9002"
+REQUESTS_HTTP_PROXY_AUTH="localhost:9003"
+REQUESTS_HTTP_PROXY_AUTH_USER="test"
+REQUESTS_HTTP_PROXY_AUTH_PASS="pass"
```
jrfnl, line 94: Considering the changes to the Composer script to use `@putenv` (in commit fb62d2b08c46dedcb843240a1f52da7ef7a8092f), so this be read out from env if available ?

scripts/start-test-environment.sh
```diff
@@ -0,0 +1,125 @@
...
+    return 1 2>/dev/null || exit 1
+fi
+
+echo "Test environment is ready!"
+echo "Environment variables set:"
+echo "REQUESTS_TEST_HOST_HTTP=localhost:8080"
+echo "REQUESTS_HTTP_PROXY=localhost:9002"
+echo "REQUESTS_HTTP_PROXY_AUTH=localhost:9003"
+echo "REQUESTS_HTTP_PROXY_AUTH_USER=test"
+echo "REQUESTS_HTTP_PROXY_AUTH_PASS=pass"
```
jrfnl, line 123: Should this be using the variables instead of having the values hard-coded in the `echo` ?

composer.json
```diff
@@ -87,25 +87,45 @@
...
+      "./scripts/start-test-environment.sh"
+    ],
+    "testserver:stop": [
+      "./scripts/stop-test-environment.sh"
+    ],
+    "test:withserver": [
+      "@putenv REQUESTS_HTTP_PROXY=localhost:9002",
+      "@putenv REQUESTS_HTTP_PROXY_AUTH=localhost:9003",
+      "@putenv REQUESTS_HTTP_PROXY_AUTH_USER=test",
+      "@putenv REQUESTS_HTTP_PROXY_AUTH_PASS=pass",
```
jrfnl, line 105: Could these be moved to a separate Composer script which is `@` referenced here ? (to avoid the duplication in the `coverage:withserver` script)

PR #918, opened 2024-12-23
Add support for HTTP2 and HTTP3

pprkut:
This adds generic support to make HTTP2 and HTTP3 requests and parse the responses correctly.

## Pull Request Type

This is a:
- [ ] Bug fix
- [x] New feature

## Context
The `protocol_version` config option now supports the values `2.0` and `3.0` to issue HTTP2 or HTTP3 requests respectively.

## Quality assurance

## Documentation

For new features:

### DIFF
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -749,7 +749,7 @@ class Requests {
 		// Unfold headers (replace [CRLF] 1*( SP | HT ) with SP) as per RFC 2616 (section 2.2)
 		$headers = preg_replace('/\n[ \t]/', ' ', $headers);
 		$headers = explode("\n", $headers);
-		preg_match('#^HTTP/(1\.\d)[ \t]+(\d+)#i', array_shift($headers), $matches);
+		preg_match('#^HTTP/(1\.\d|2|3)[ \t]+(\d+)#i', array_shift($headers), $matches);
 		if (empty($matches)) {
 			throw new Exception('Response could not be parsed', 'noversion', $headers);
 		}
--- a/src/Transport/Curl.php
+++ b/src/Transport/Curl.php
@@ -393,11 +393,11 @@ final class Curl implements Transport {
 		 * add as much as a second to the time it takes for cURL to perform a request. To
 		 * prevent this, we need to set an empty "Expect" header. To match the behaviour of
 		 * Guzzle, we'll add the empty header to requests that are smaller than 1 MB and use
-		 * HTTP/1.1.
+		 * a protocol version newer than HTTP/1.0.
 		 *
 		 * https://curl.se/mail/lib-2017-07/0013.html
 		 */
-		if (!isset($headers['Expect']) && $options['protocol_version'] === 1.1) {
+		if (!isset($headers['Expect']) && $options['protocol_version'] > 1.0) {
 			$headers['Expect'] = $this->get_expect_header($data);
 		}
 
@@ -465,10 +465,19 @@ final class Curl implements Transport {
 			curl_setopt($this->handle, CURLOPT_HTTPHEADER, $headers);
 		}
 
-		if ($options['protocol_version'] === 1.1) {
-			curl_setopt($this->handle, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);
-		} else {
-			curl_setopt($this->handle, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0);
+		switch ($options['protocol_version']) {
+			case 3.0:
+				// The CURL_HTTP_VERSION_3 constant is only available from PHP 8.4
+				curl_setopt($this->handle, CURLOPT_HTTP_VERSION, 30);
+				break;
+			case 2.0:
+				curl_setopt($this->handle, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
+				break;
+			case 1.1:
+				curl_setopt($this->handle, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);
+				break;
+			default:
+				curl_setopt($this->handle, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0);
 		}
 
 		if ($options['blocking'] === true) {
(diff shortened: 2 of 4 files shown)


schlessera:
Hi @pprkut ,

Thanks for your work on this!

Right now, it is not in a mergeable state, as the testing that is included is not sufficient. The testing fails to verify that this solves the issue it is meant to solve, so it is unclear whether this actually makes HTTP2 and HTTP3 work. From what I've gathered so far, it only allows for the connected server to specify to use HTTP2 or HTTP3 without breaking early, but then it is unclear whether it will work or break on a bad protocol because there might still be pieces missing to actually get the communication to work flawlessly.

It also creates a discrepancy between the curl and fsockopen transports. While we can debate whether we should look into HTTP2/HTTP3 support for fsockopen, at the very least we should have a capability check in place to let a user of the library detect whether they can make use of them or not with the protocol(s) supported on the current platform.

Also, we'll need our test server (=> https://github.com/RequestsPHP/test-server) and our test proxies to allow for proper testing of these.

I'm happy to help work on this and move it forward, but I have to admit I'm not deep enough into the technical details at this point to know the full details of how we can set this up for proper testing. is this something you are able and willing to help with?

pprkut:
Thanks @schlessera for taking a look. I totally agree that there is more work left here. What I **can** say is that we needed this change to talk to Google's HTTP2 APIs and that seems to work (though I can't find concrete evidence in the logged headers that we're actually using HTTP2).

This is a critical feature for us, so I'll be happy to spend some more time on it to get it upstreamed :slightly_smiling_face:

pprkut:
PHP's built-in web server seems to only support HTTP 1.1 (not even 1.0), based on https://github.com/php/php-src/blob/master/sapi/cli/php_cli_server.c#L719
I don't think we can fit HTTP2/3 support into the existing test server :confused:

schlessera:
Thanks for the additional context, @pprkut, and I'm happy to hear you can invest some more time into this! 😊

Yes, so what we really need to get properly started is a strategy on how to test this reliably so we can say with certitude that the entire round-trip is fully working with HTTP2/HTTP3 - sending a header claiming it is supported is not enough.

Re. the existing test server, that is just one way of testing that we have available. Right now, we also have dockerized setup (in https://github.com/RequestsPHP/test-server/tree/feature/docker-setup branch) that is being run on https://render.com/ (=> https://requests-test-server.onrender.com/). And I'm happy to add additional infrastructure or services if that's what is needed to get reliable tests going. But I'm unsure what the best approach is here without more research.

Btw, just fetching https://requests-test-server.onrender.com/get from the browser seems to return it in HTTP/2.

pprkut:
There's `CURLINFO_HTTP_VERSION` which returns the HTTP version used for the last request, which should be certain enough I think. The downside is that that is only available with curl >= 7.50.0.
HTTP2 support was added in curl 7.33.0.

I made some sample requests to the test server you mentioned (as well as https://www.google.com) and it seems to support both HTTP2 and HTTP3 just fine. This is the test script I used:

```php
<?php

$ch = curl_init();

curl_setopt($ch, CURLOPT_URL, 'https://requests-test-server.onrender.com/get');
// curl_setopt($ch, CURLOPT_URL, 'https://www.google.com');
curl_setopt($ch, CURLOPT_RETURNTRANSFER, TRUE);
curl_setopt($ch, CURLOPT_HTTP_VERSION, 31);

$page = curl_exec($ch);

echo "CURL_HTTP_VERSION_1_0: " . CURL_HTTP_VERSION_1_0 . "\n";
echo "CURL_HTTP_VERSION_1_1: " . CURL_HTTP_VERSION_1_1 . "\n";
echo "CURL_HTTP_VERSION_2: " . CURL_HTTP_VERSION_2 . "\n";
echo "CURL_HTTP_VERSION_2TLS: " . CURL_HTTP_VERSION_2TLS . "\n";
echo "CURL_HTTP_VERSION_2_0: " . CURL_HTTP_VERSION_2_0 . "\n";
echo "CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE: " . CURL_HTTP_VERSION_2_PRIOR_KNOWLEDGE . "\n";
echo "CURL_HTTP_VERSION_3: " . 30 . "\n";
echo "CURL_HTTP_VERSION_3ONLY: " . 31 . "\n";

$info = curl_getinfo($ch, CURLINFO_HTTP_VERSION);

var_dump($info);

curl_close($ch);
```

masakielastic:
Thanks for working on this. I came across this PR while investigating HTTP/2 support in WP-CLI.

While looking into the issue, I found a related but narrower question: Requests currently forces `CURL_HTTP_VERSION_1_1`, preventing libcurl from using its native protocol negotiation.

I've opened #1087 to discuss that separately, so as not to expand the scope of this PR.

I also think separating the concerns may help with testing: HTTP/2 and HTTP/3 require substantially different transport and test infrastructure, while native cURL negotiation could potentially be introduced and tested independently first.

Further discussion about transport-native negotiation can continue in #1087.

masakielastic:
I think HTTP/2 and HTTP/3 should probably be considered separately here.

HTTP/2 in pure PHP is technically possible. For example, PSL now has an H2 implementation covering HTTP/2 framing, HPACK, flow control and multiplexing, and its HTTP client supports HTTP/2 over TLS/ALPN and h2c.

That said, I would see PSL mainly as proof that this is feasible rather than something Requests could directly reuse. PSL's current H2 stack requires PHP 8.4+, while Requests still supports PHP 5.6, and implementing and maintaining an equivalent HTTP/2 stack inside Requests would be a substantial increase in scope.

HTTP/3 seems like a different category entirely. It is not just another HTTP framing layer over the existing socket transport; it requires QUIC over UDP, TLS 1.3 integration, packet protection, loss recovery, stream management, etc.

For comparison, Python's `aioquic` provides a QUIC stack, a minimal TLS 1.3 implementation and an HTTP/3 stack, and still relies on cryptographic/OpenSSL tooling and extensive interoperability testing.

So my inclination would be:

* HTTP/2 over a pure-PHP transport is theoretically feasible, but probably beyond what Requests needs to implement itself right now.
* HTTP/3 should probably be treated as a separate QUIC-capable transport rather than an extension of the existing `fsockopen` transport.
* For the immediate HTTP/2 use case, allowing the cURL transport to use libcurl's native protocol negotiation seems much smaller and easier to maintain.

This would also let Requests benefit from HTTP/2 without making pure-PHP HTTP/2/3 support a prerequisite.

PR #925, opened 2025-01-22
GH Actions: add actionlint job

jrfnl:
While GH Actions will fail runs of scripts which contain syntax errors, it isn't always that straight forward to see what's wrong.

The `actionlint` package helps with that by providing more informative error messages. Actionlint also checks for a number of best practice to help keep the GH Actions scripts secure and working.

Ref: https://github.com/rhysd/actionlint

### DIFF
--- a/.github/workflows/cs.yml
+++ b/.github/workflows/cs.yml
@@ -14,6 +14,28 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  actionlint: #----------------------------------------------------------------------
+    name: 'Check GHA workflows'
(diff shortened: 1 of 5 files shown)


schlessera (approved):

schlessera (approved):

jrfnl:
Note: the shellcheck code which is being ignored is about combining multiple outputs, but this is not great for documentation of what's being done nor for debugging, which is why I elected to ignore that code.

https://github.com/WordPress/Requests/actions/runs/12886164407/job/35926141799

PR #933, opened 2025-06-17
Requests.php: always set body text from response

granthony:
This fixes a bug in which response body text satisfying PHP "empty()" is not returned to the caller.

As substr() always returns a string, there is no need for the guard condition here.  This has been removed accordingly, resolving the issue.

## Pull Request Type

- [ ] I have checked there is no other PR open for the same change.

This is a:
- [X] Bug fix
- [ ] New feature

## Context

In dealing with a third-party API we found that empty responses were sometimes returned when we were expecting the non-empty response "0".  Requests turned out to be at fault.

Expected behaviour: the response string "0" is returned.
Observed behaviour: no response is returned.

## Detailed Description

In limited circumstances, the response body text is not correctly returned to the caller.  PHP empty() treats some unintuitive things as empty, including the string "0".

## Quality assurance

- [ ] This change does NOT contain a breaking change (fix or feature that would cause existing functionality to change).

This is a breaking change but is likely to be of minimal impact.

A unit test could be added for requests returning the response string "0".

## Documentation

For new features:

### DIFF
--- a/src/Requests.php
+++ b/src/Requests.php
@@ -739,9 +739,7 @@ class Requests {
 			$headers = substr($return->raw, 0, $pos);
 			// Headers will always be separated from the body by two new lines - `\n\r\n\r`.
 			$body = substr($return->raw, $pos + 4);
-			if (!empty($body)) {
-				$return->body = $body;
-			}
+			$return->body = $body;
 		}
 
 		// Pretend CRLF = LF for compatibility (RFC 2616, section 19.3)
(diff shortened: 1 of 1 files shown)


jrfnl (changes requested):
@granthony Thanks for this PR and your willingness to contribute. However, this "bug fix" introduces a new bug.

> As substr() always returns a string, there is no need for the guard condition here. This has been removed accordingly, resolving the issue.

This is true for PHP >= 8.0, but not for PHP < 8.0, in which case, `substr()` could return `false`. And `false` is not a valid value for the `Response::$body`.

https://www.php.net/manual/en/function.substr.php#refsect1-function.substr-changelog

granthony:
My mistake, thank you.  A strict comparison against false should do then.

jrfnl:
> My mistake, thank you. A strict comparison against false should do then.

As a rule of thumb, it's better to check for what you want, than what you _don't_ want, so maybe the check should be a `is_string()` instead ?

Aside from that, I believe this change should be covered by a test. Could you please add one ?

PR #947, opened 2025-09-01
PHP 8.5: Prevent deprecation notices for curl_close

TobiasBg:
`curl_close` is deprecated in PHP 8.5+, and hasn't been doing anything since PHP 8.0.

To prevent deprecation warnings it should therefore be called on older versions of PHP only.

See https://wiki.php.net/rfc/deprecations_php_8_5#deprecate_no-op_functions_from_the_resource_to_object_conversion.

## Pull Request Type

This is a:
- [ ] Bug fix
- [ ] New feature

## Quality assurance

## Documentation

For new features:

### DIFF
--- a/src/Transport/Curl.php
+++ b/src/Transport/Curl.php
@@ -323,7 +323,9 @@ final class Curl implements Transport {
 				}
 
 				curl_multi_remove_handle($multihandle, $done['handle']);
-				curl_close($done['handle']);
+				if (is_resource($done['handle'])) {
+					curl_close($done['handle']);
+				}
 
 				if (!is_string($responses[$key])) {
 					$options['hooks']->dispatch('multiple.request.complete', [&$responses[$key], $key]);


jrfnl (changes requested):
@TobiasBg Thanks for this PR.

The change on line 326 looks valid, but the other two changes are redundant. The conditions already check if `$this->handle` is a resource and that condition would return `false` as of PHP 8.0 (as it would be an object, not a resource), so adding an additional version comparison will not make a difference...

jrfnl:
P.s.: probably better to change the version comparison on line 326 to a check for `is_resource()` as well for stability (if that is not checked earlier in the logic, in which case, the change on line 326 would be redundant too....)

TobiasBg:
Oh, thanks @jrfnl! Good to know about the `resource`/`object` switch.

I have updated the PR, to only leave that one line, with an `is_resource` check.

jrfnl:
> Good to know about the `resource`/`object` switch.

It helps to understand a deprecation before addressing it... Not sure if you pulled similar PRs elsewhere, but if so, you may want to update those too...

TobiasBg:
Yes, checked those already. Thanks.

I think that including the PHP version check can be helpful in the future, when support for older versions of PHP is dropped in projects.

jrfnl:
> I think that including the PHP version check can be helpful in the future, when support for older versions of PHP is dropped in projects.

I don't agree. That's what comments are for, or one could open an issue with a tasklist as a reminder of things which would need to be done in the future, but the code itself should be based on what makes it most stable. Considering that people can compile custom versions of PHP with different version of extensions (no matter how unlikely it is for anyone to do that in this case), the `is_resource()` check creates the most stable code.

TobiasBg:
Great points and arguments! Changed my mind :-) Thanks!

schlessera (approved):

schlessera:
Thanks for your PR, @TobiasBg !

Issue #966, opened 2025-11-20
Investigate failing test for revoked certificate


jrfnl:
PR #967 has marked the `Transport\Curl\BaseTestCase::testRevokedHTTPS()` test(s) as skipped as the tests were failing.

First off, the original URL we were using - `https://testssl-revoked.disig.sk/index.en.html` - no longer worked.

This was causing the tests to fail like so:
```
2) WpOrg\Requests\Tests\Transport\Curl\CurlTest::testRevokedHTTPS
Failed asserting that exception message 'cURL error 60: SSL certificate problem: self-signed certificate' contains 'certificate subject name'.

4) WpOrg\Requests\Tests\Transport\Fsockopen\FsockopenTest::testRevokedHTTPS
Failed asserting that exception message 'stream_socket_client(): SSL operation failed with code 1. OpenSSL Error messages:
error:0A000086:SSL routines::certificate verify failed
stream_socket_client(): Failed to enable crypto
stream_socket_client(): unable to connect to ssl://testssl-revoked.disig.sk:443 (Unknown error)' contains 'SSL certificate did not match the requested domain name'.
```

This was fixed in PR #967 by switching the URL used in the tests to the `https://revoked.badssl.com/` URL.

However, now the tests are still failing:

```
There were 2 failures:

1) WpOrg\Requests\Tests\Transport\Curl\CurlTest::testRevokedHTTPS
Failed asserting that exception of type "WpOrg\Requests\Exception" is thrown.

2) WpOrg\Requests\Tests\Transport\Fsockopen\FsockopenTest::testRevokedHTTPS
Failed asserting that exception of type "WpOrg\Requests\Exception" is thrown.
```

This indicates there is an underlying problem with the Request code and/or the test.

According to some unknown source:

> Certificate revocation checking is not enabled by default in PHP's cURL or fsockopen. The certificate is technically valid (not expired, proper domain), it's just been revoked, but that's only checked if you explicitly enable OCSP/CRL checking.

Tasks: 
1. The above statement about PHP needs to be confirmed.
2. We'd then need to decide how Requests should handle this. (ignore the revocation as it apparently does now, or throw an exception, which appears to be the behaviour the original design planned for)
3. Based on what direction is chosen, the code and/or the tests need updating.

PR #967, opened 2025-11-20
Fix tests for expired and revoked SSL certificates

schlessera:
The tests for expired or revoked SSL certificates were failing for both Curl and FsockOpen because the test site we had been using has changed (https://testssl-expire.disig.sk/index.en.html).

This PR switches the test site to use expired/revoked.badssl.com, which is more of an industry standard.

The exception message to expect needed to be changed as well to make the expired test pass.

The revoked test has been set to be skipped, as it turns out that revocation checking is disabled by default for PHP's cURL.

See #966


jrfnl (approved):
Thanks @schlessera !

PR #999, opened 2025-12-01
GH Actions/update-website: stop submitting test PRs on workflow update

jrfnl:
## Pull Request Type

## Detailed Description

Currently, the `update-website` workflow doesn't deploy the website straight away. Instead it opens a PR with the files which will be changed to allow for review of the website changes before deploying the updated website (by merging the PR).

To that end, the workflow is run:
* Whenever a tag is created.
* When a pull request changes any of the files directly involved in updating the website.
* When manually requested.

What with the frequent Dependabot updates, this has become quite noisy for the "pull request which changes the workflow" trigger.

To that end, this commit blocks the opening of the PR when the workflow/doc building files are changed, while still running the rest of the workflow to verify that the workflow wasn't broken by the changes.

It also introduces a new step to create the branch holding the changes in the local environment.
This was previously done via the "create PR" step, which will now be skipped, while we still need the branch for verifying Ruby can build the GH Pages site.

As the branch should now be checked-out already before we even get to the part of the workflow verifying the site can be build via GH Pages, we can also remove the separate `actions/checkout` step.

### DIFF
--- a/.github/workflows/update-website.yml
+++ b/.github/workflows/update-website.yml
@@ -125,6 +125,11 @@ jobs:
           name: website-updates
           path: artifacts
 
+      - name: Create the new branch
+        env:
+          REF_NAME: feature/auto-ghpages-update-${{ steps.get_pr_info.outputs.REF }}
+        run: git checkout -b "$REF_NAME"
+
       # Different version of phpDocumentor may add/remove files for CSS/JS etc.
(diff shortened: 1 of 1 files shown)


jrfnl:
Sorry for the commit noise, hoping things are finally correct.

schlessera (approved):

PR #1015, opened 2025-12-12
Add Host Bindings feature

schlessera:
## Pull Request Type

This is a:
- [ ] Bug fix
- [x] New feature

## Context

This PR adds support for binding specific hostnames to specific IP addresses, bypassing DNS resolution. 

This is useful for:
- Testing against specific server instances
- Load balancing and failover scenarios
- Connecting to servers via specific IPs while preserving the original hostname in the `Host` header
- Development environments where DNS may not be configured
- Preventing rebinding attacks

### New `HostBindings` utility class

A value object that validates and stores hostname-to-IP mappings. Key features:

* **Security by default**: IP addresses are validated using `filter_var(FILTER_VALIDATE_IP)` to prevent hostname injection attacks. Accepts both IPv4 and IPv6 addresses (including private/localhost ranges).
* **Opt-out mechanism**: For pre-validated inputs or non-standard formats, validation can be skipped via `HostBindings::SKIP_IP_VALIDATION` (use with caution).
* **Normalization**: Whitespace is trimmed from IP addresses.

Methods:
* `has_host($host)` - Check if a host has a mapping
* `get_first_ip_for_host($host)` - Get the first IP for a host (throws `UnknownHost` or `MissingIpAddress` if not available)
* `get_all_ips_for_host($host)` - Get all IPs for a host (throws `UnknownHost` if not found)

### New exceptions

* `UnknownHost` - Thrown when requesting a host that isn't in the bindings
* `MissingIpAddress` - Thrown when a host has no IP addresses configured

### New `HOST_BINDINGS` capability

Added to the `Capability` interface to allow transport capability testing.

### Transport integration

**Curl transport** (`Transport\Curl`):
* Uses `CURLOPT_CONNECT_TO` (cURL 7.49.0+) as the preferred method
* Falls back to `CURLOPT_RESOLVE` (cURL 7.21.3+) for older cURL versions
* Falls back to URL rewriting for HTTP (not HTTPS) on very old cURL versions
* Properly handles IPv6 addresses by wrapping them in square brackets
* Preserves the original hostname in the `Host` header

**Fsockopen transport** (`Transport\Fsockopen`):
* Replaces the connection host with the mapped IP address
* Original hostname is preserved in the `Host` header

### Usage

The `host_bindings` option accepts either an array (with automatic IP validation) or a pre-constructed `HostBindings` object:

```php
// Using an array (IPs are validated)
$response = Requests::get('https://example.com/api', [], [
    'host_bindings' => [
        'example.com' => ['93.184.216.34', '2606:2800:220:1:248:1893:25c8:1946'],
    ],
]);

// Using a HostBindings object (for advanced control)
$bindings = new HostBindings(
    ['example.com' => ['custom-value']],
    HostBindings::SKIP_IP_VALIDATION
);
$response = Requests::get('https://example.com/api', [], [
    'host_bindings' => $bindings,
]);
```

### Includes

* Comprehensive unit tests for `HostBindings` class (constructor validation, all methods)
* Unit tests for new exceptions
* Integration tests in `Transport\BaseTestCase` covering both transports
* Documentation in `Requests::request()` docblock


jrfnl:
Prelim remark: how about adding that code example (and maybe some more text) to the documentation for the website ? As in: either add or update one of the existing markdown files in the `docs` directory ?

jrfnl:
@schlessera Would you mind also having a look at what's happening with the code coverage ? I would not expect a new feature, which is described as "fully tested" to reduce code coverage.

src/Exception/MissingIpAddress.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * Requests for PHP, an HTTP library.
+ *
+ * @copyright 2012-2023 Requests Contributors
```
jrfnl, line 5:
```suggestion
 * @copyright 2012-2026 Requests Contributors
```

(here and elsewhere - or should we do this codebase wide after this PR has been merged ?)

src/Exception/MissingIpAddress.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace WpOrg\Requests\Exception;
+
+use RangeException;
+
+/**
+ * Exception for a missing IP address in the host bindings.
+ *
+ * @package Requests\Exceptions
+ * @since   2.x.x
```
jrfnl, line 18:
```suggestion
 * @since   2.1.0
```

src/Exception/MissingIpAddress.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+final class MissingIpAddress extends RangeException {
+
+	/**
+	 * Instantiate a MissingIpAddress exception for a missing IP address in the host bindings.
+	 *
+	 * @param string $host Host that was requested.
+	 * @return self
```
jrfnl, line 26:
```suggestion
	 * @return \WpOrg\Requests\Exception\MissingIpAddress
```

src/Exception/UnknownHost.php
```diff
@@ -0,0 +1,34 @@
...
+ * @since   2.x.x
+ */
+final class UnknownHost extends InvalidArgumentException {
+
+	/**
+	 * Instantiate an UnknownHost exception for an unknown host that was
+	 * requested via HostBindings.
+	 *
+	 * @param string $host Unknown host that was requested.
+	 * @return self
```
jrfnl, line 27:
```suggestion
	 * @return \WpOrg\Requests\Exception\UnknownHost
```

src/Exception/UnknownHost.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace WpOrg\Requests\Exception;
+
+use InvalidArgumentException;
+
+/**
+ * Exception for an unknown host being requested via HostBindings.
+ *
+ * @package Requests\Exceptions
+ * @since   2.x.x
```
jrfnl, line 18:
```suggestion
 * @since   2.1.0
```

src/Exception/MissingIpAddress.php
```diff
@@ -0,0 +1,33 @@
...
+	 */
+	public static function for_host($host) {
```
jrfnl, line 28:
```suggestion
	public static function create($host) {
```

Reason for this suggestion: in other Exceptions introduced over the past few years, we've used the pattern of using `static` initializers called `create()`.

src/Exception/UnknownHost.php
```diff
@@ -0,0 +1,34 @@
...
+final class UnknownHost extends InvalidArgumentException {
+
+	/**
+	 * Instantiate an UnknownHost exception for an unknown host that was
+	 * requested via HostBindings.
+	 *
+	 * @param string $host Unknown host that was requested.
+	 * @return self
+	 */
+	public static function for_host($host) {
```
jrfnl, line 29:
```suggestion
	public static function create($host) {
```

Reason for this suggestion: in other Exceptions introduced over the past few years, we've used the pattern of using `static` initializers called `create()`.

src/Requests.php
```diff
@@ -460,9 +467,23 @@ public static function request($url, $headers = [], $data = [], $type = self::GE
...
+			$need_ssl = (stripos($url, 'https://') === 0);
+
+			$need_host_bindings =
+				array_key_exists(Capability::HOST_BINDINGS, $options)
+				&&
+				(
+					is_array($options[Capability::HOST_BINDINGS])
+					||
+					$options[Capability::HOST_BINDINGS] instanceof HostBindings
+				);
```
jrfnl, line 479:
```suggestion
			$need_host_bindings = array_key_exists(Capability::HOST_BINDINGS, $options)
				&& (is_array($options[Capability::HOST_BINDINGS])
					|| $options[Capability::HOST_BINDINGS] instanceof HostBindings
				);
```

src/Transport/Curl.php
```diff
@@ -385,8 +387,10 @@ public function &get_subrequest_handle($url, $headers, $data, $options) {
 	private function setup_handle($url, $headers, $data, $options) {
 		$options['hooks']->dispatch('curl.before_request', [&$this->handle]);
 
+		$case_insensitive_headers = new CaseInsensitiveDictionary($headers);
+
 		// Force closing the connection for old versions of cURL (<7.22).
-		if (!isset($headers['Connection'])) {
+		if (!isset($case_insensitive_headers['Connection'])) {
```
jrfnl, line 393:
```suggestion
		if (!isset($case_insensitive_headers['connection'])) {
```

If you are making it case-insensitive.... ?

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+use WpOrg\Requests\Exception\UnknownHost;
+
+/**
+ * Value object to handle host bindings that are provided via the $options array.
+ *
+ * Each mapping maps a host URL string to an array of one or more IPv4/IPv6 addresses.
+ * IP addresses are validated by default to prevent hostname injection attacks.
+ *
+ * Security Note:
+ * - IP addresses are validated using filter_var(FILTER_VALIDATE_IP) by default
```
jrfnl, line 23: If this is true, then the `filter` extension becomes a new required dependency for the Requests library for anyone wanting to use the host binding feature.

So.... this will need to be annotated in the `composer.json` `require` or `suggest` section and we will also need a check for `filter_var()` being available before making that function call.

I wonder how problematic this will turn out to be for WordPress ? Making a PHP extension required is not something they are keen on in my experience....

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+ * - Accepts valid IPv4 and IPv6 addresses (including private/localhost)
+ * - Rejects hostnames, URLs, and invalid formats
+ * - Validation can be disabled for pre-validated inputs (use with caution)
+ *
+ * @package Requests\Utilities
+ * @since   2.x.x
```
jrfnl, line 29:
```suggestion
 * @since   2.1.0
```

(here and everywhere else in this class - though I wonder whether the `@since` tags on the components of the class are really needed - the class is being introduced in `2.1.0`, shouldn't we only add component level `@since` tags for those components introduced _after_ `2.1.0` ?)

src/Capability.php
```diff
@@ -19,6 +19,13 @@
  */
 interface Capability {
 
+	/**
+	 * Support for mapping specific hosts to specific IP addresses.
+	 *
```
jrfnl, line 24:
```suggestion
	 *
	 * @since 2.1.0
	 *
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+ */
+final class HostBindings {
+
+	/**
+	 * Enable IP address validation in the constructor (default, recommended).
+	 *
+	 * @since 2.x.x
```
jrfnl, line 36:
```suggestion
```

(see remark above)

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	 *
+	 * WARNING: Only use if you have already validated the IP addresses yourself,
+	 * or need to use non-standard address formats. Skipping validation may allow
+	 * hostname injection attacks if untrusted data is used.
+	 *
+	 * @since 2.x.x
+	 *
+	 * @var bool
+	 */
+	const SKIP_IP_VALIDATION = false;
```
jrfnl: How can this feature be used ? This is a `final` class, so this constant can not be overloaded and as it is a constant, it cannot be changed at runtime either.

What am I missing ?

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+final class HostBindings {
+
+	/**
+	 * Enable IP address validation in the constructor (default, recommended).
+	 *
+	 * @since 2.x.x
+	 *
+	 * @var bool
+	 */
+	const VALIDATE_IPS = true;
```
jrfnl, line 40: As discussed - it took me a little to understand that this constant - and the `SKIP_IP_VALIDATION` one below - is supposed to be a flag to be used as a value to be passed to the class constructor.

I think the following changes would make this more obvious:
* Rename the constants so they both use the same prefix. Maybe `IP_VALIDATION` and `IP_VALIDATION_SKIP` ? But more descriptive names are welcome.
* Use the word `flag` in the docblock descriptions of the constants and indicate in the description that these are supposed to be used a values to be passed to the constructor.

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	/**
+	 * Construct HostBindings object.
+	 *
+	 * @since 2.x.x
+	 *
+	 * @param array<string, array<string>> $host_bindings Host to IP address mappings.
+	 * @param bool                          $validate_ips  Whether to validate IP address formats.
+	 *                                                      Default: true (recommended).
+	 *                                                      Set to false ONLY if you have already
+	 *                                                      validated the IP addresses yourself.
```
jrfnl, line 71: The below is just a formatting fix. The more important remark here is that the parameter description should indicate that this expects one of the two class constants to be passed. Not a plain boolean.

```suggestion
	 * @param bool                         $validate_ips  Whether to validate IP address formats.
	 *                                                    Default: true (recommended).
	 *                                                    Set to false ONLY if you have already
	 *                                                    validated the IP addresses yourself.
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	 *
+	 * @throws InvalidArgument If parameters are invalid or IPs fail validation.
```
jrfnl, line 73:
```suggestion
	 * @throws \WpOrg\Requests\Exception\InvalidArgument If parameters are invalid or IPs fail validation.
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	 * @param array<string,array<string>> $host_bindings Host bindings to validate.
+	 * @param bool                        $validate_ips  Whether to validate IP formats.
+	 *
+	 * @return array<string,array<string>> Validated and normalized array of host bindings.
+	 *
+	 * @throws \WpOrg\Requests\Exception\InvalidArgument If validation fails.
+	 */
+	private function validate($host_bindings, $validate_ips) {
+		$validated = [];
+
```
jrfnl, line 101: Would bowing out early if `$host_bindings` is an empty array make sense here ?

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+		$this->host_bindings = $this->validate($host_bindings, $validate_ips);
+	}
+
+	/**
+	 * Validate and normalize a passed-in host bindings array.
+	 *
+	 * @since 2.x.x
+	 *
+	 * @param array<string,array<string>> $host_bindings Host bindings to validate.
+	 * @param bool                        $validate_ips  Whether to validate IP formats.
```
jrfnl, line 93:
```suggestion
	 * @param array<string, array<string>> $host_bindings Host bindings to validate.
	 * @param bool                         $validate_ips  Whether to validate IP formats.
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	 *
+	 * @return array<string,array<string>> Validated and normalized array of host bindings.
```
jrfnl, line 95:
```suggestion
	 * @return array<string, array<string>> Validated and normalized array of host bindings.
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+			// Validate host key.
+			if (is_string($host) === false || $host === '') {
+				throw InvalidArgument::create(1, '$host_bindings (host)', 'non-empty string', gettype($host));
+			}
+
+			// Validate IPs array.
+			if (is_array($ips) === false) {
+				throw InvalidArgument::create(1, '$host_bindings (ip address array)', 'array', gettype($ips));
+			}
+
```
jrfnl, line 112: Just wondering: should we bow out early if the `$ips` array is empty ? Or maybe that should throw an `InvalidArgument` exception ?

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	 * @var bool
+	 */
+	const SKIP_IP_VALIDATION = false;
+
+	/**
+	 * Array of host bindings.
+	 *
+	 * @var array<string, array<string>>
+	 */
+	private $host_bindings;
```
jrfnl, line 60: Unfinished brainfart: should we protect this class against serialization ? If an instance of this class would be stored in serialized form, unserialized and reinstated, I have a niggly feeling we might run the risk of unvalidated hosts being injected in the host binding.

Maybe I'm searching for problems where there are none, but I think we need to think this through a bit more.

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+						'$host_bindings (ip address value)',
+						'non-empty IP address',
+						'empty string provided'
+					);
+				}
+
+				// Conditionally validate IP format.
+				if ($validate_ips === true) {
+					// Validate both IPv4 and IPv6, including private/localhost ranges.
+					if (filter_var($ip_trimmed, FILTER_VALIDATE_IP) === false) {
```
jrfnl, line 135: Genuine question: is the `FILTER_VALIDATE_IP` flag alone sufficient ? Or should this be combined with any of the related flags ?

https://www.php.net/manual/en/filter.constants.php#constant.filter-validate-ip

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	/**
+	 * Get the first IP address mapping in the host bindings for a particular host.
+	 *
+	 * This throws an exception if the requested host does not have a mapping.
+	 * This also throws an exception if the requested host has no IP addresses available.
+	 *
+	 * @param string $host Host to request a mapping for.
+	 * @return string IP address mapping for the host.
+	 * @throws UnknownHost If the requested host does not have a mapping.
+	 * @throws MissingIpAddress If the requested host has no IP address available.
```
jrfnl, line 177:
```suggestion
	 *
	 * @throws \WpOrg\Requests\Exception\UnknownHost      If the requested host does not have a mapping.
	 * @throws \WpOrg\Requests\Exception\MissingIpAddress If the requested host has no IP address available.
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+		}
+
+		return array_key_exists($host, $this->host_bindings);
+	}
+
+	/**
+	 * Get the first IP address mapping in the host bindings for a particular host.
+	 *
+	 * This throws an exception if the requested host does not have a mapping.
+	 * This also throws an exception if the requested host has no IP addresses available.
```
jrfnl, line 172: This looks redundant, what with the `@throws` tags containing the same info.

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+	/**
+	 * Get all IP address mappings in the host bindings for a particular host.
+	 *
+	 * This throws an exception if the requested host does not have a mapping.
+	 * Contrary to get_first_ip_for_host(), this method does not throw an exception
+	 * if the host has no IP addresses available. It will simply return an empty array.
+	 *
+	 * @param string $host Host to request a mapping for.
+	 * @return array<string> IP address mappings for the host.
+	 * @throws UnknownHost If the requested host does not have a mapping.
```
jrfnl, line 200:
```suggestion
	 *
	 * @throws \WpOrg\Requests\Exception\UnknownHost If the requested host does not have a mapping.
```

src/Utility/HostBindings.php
```diff
@@ -0,0 +1,209 @@
...
+
+		return $this->host_bindings[$host][0];
+	}
+
+	/**
+	 * Get all IP address mappings in the host bindings for a particular host.
+	 *
+	 * This throws an exception if the requested host does not have a mapping.
+	 * Contrary to get_first_ip_for_host(), this method does not throw an exception
+	 * if the host has no IP addresses available. It will simply return an empty array.
```
jrfnl, line 196: Just wondering - why still throw an exception if the host doesn't have a mapping ? Why not return an empty array in that case too ?

src/Transport/Curl.php
```diff
@@ -418,6 +420,99 @@ private function setup_handle($url, $headers, $data, $options) {
...
+			$exec_ip = strpos($exec_ip, ':') === false ? $exec_ip : "[{$exec_ip}]";
+
+			// @TODO: Extract connect_to/resolve handling into separate method.
+			if (defined('CURLOPT_CONNECT_TO')) {
+				$connect_to_string = "{$host}:{$normalized_port}:{$exec_ip}:{$normalized_port}";
+				// phpcs:ignore PHPCompatibility.Constants.NewConstants.curlopt_connecttoFound
+				curl_setopt($this->handle, CURLOPT_CONNECT_TO, [$connect_to_string]); // phpcs:ignore PHPCompatibility.Constants.NewConstants.curlopt_connect_toFound
+				// Fallback branches for cURL < 7.49.0 which lacks CURLOPT_CONNECT_TO.
+				// Cannot be reached in tests as CURLOPT_CONNECT_TO is always defined.
+				// @codeCoverageIgnoreStart -- The else is never reached in CI because CURLOPT_CONNECT_TO is always defined.
```
```suggestion
				// Fallback branches for cURL < 7.49.0 which lacks CURLOPT_CONNECT_TO.
				// Cannot be reached in tests as CURLOPT_CONNECT_TO is always defined.
				// The `else` is never reached in CI because CURLOPT_CONNECT_TO is always defined.
				// @codeCoverageIgnoreStart
```

Might need testing, but I have a niggly feeling that the PHPUnit code coverage annotations do not support adding a comment behind the annotation.
That might explain why the comment isn't working as intended when viewing the CodeCov report.

jrfnl (commented):
This is not a complete review (yet). Posting it anyway to allow for further iteration on this PR based on my remarks.

PR #1016, opened 2025-12-12
GH Actions/update-website: fix workflow

jrfnl:
Follow up on PR #999, which broke the workflow for submitting a PR when a new release is tagged.


schlessera (approved):

jrfnl:
Seeing PR #1017, looks like this fix worked. Will review #1017 in a moment.

PR #1060, opened 2026-07-10
Add streaming support

### LINKED ISSUE #1056: Add support for streaming response bodies
## Is your feature request related to a problem?

As part of the ongoing AI work, we want core to stream AI responses so tokens can be shown as they arrive, instead of waiting for the whole response to finish. We cannot do this today because Requests offers no readable response body you can pull from incrementally. By default it accumulates the entire body in memory before returning ([cURL](https://github.com/WordPress/Requests/blob/ca483f98a377f5cbcbd0567696f48ecd3e76623f/src/Transport/Curl.php#L584), [fsockopen](https://github.com/WordPress/Requests/blob/ca483f98a377f5cbcbd0567696f48ecd3e76623f/src/Transport/Fsockopen.php#L338)). The only ways to touch bytes mid-transfer are push-only and run entirely inside the blocking request call: the `request.progress` hook (which still buffers the whole body and cannot abort), and writing to a handle through `filename` (write-only, no readable stream). Neither yields a stream the AI SDK can pull from.

This blocks the AI client SDK. Its streaming support (WordPress/php-ai-client#255) reads the response body as a stream and pulls from it as data arrives, so it needs an HTTP client that can return a body that is read incrementally. WordPress cannot provide one today, because the HTTP client underneath buffers the response and only returns it once it is complete.

## Describe the solution you'd like

Add an opt-in way to read a response body incrementally as it arrives, without holding the whole body in memory.

The shape the SDK needs is a body that can be pulled from while the response is still streaming, so a PSR-18 adapter over `WP_Http` can expose it as a readable stream. Rough shape:

```php
$response = Requests::request($url, $headers, $data, $type, ['stream' => true]);
while (!$response->eof()) {
    $chunk = $response->read(8192); // returns data as it arrives off the connection
    // handle each chunk
}
```

When streaming is not requested, nothing changes and behavior stays exactly as it is today. This gives core the primitive it needs to extend `WP_Http` and let the AI client stream on WordPress.

## Describe alternatives you've considered

- The existing `request.progress` hook, which both transports dispatch ([cURL](https://github.com/WordPress/Requests/blob/ca483f98a377f5cbcbd0567696f48ecd3e76623f/src/Transport/Curl.php#L567), [fsockopen](https://github.com/WordPress/Requests/blob/ca483f98a377f5cbcbd0567696f48ecd3e76623f/src/Transport/Fsockopen.php#L323)), and which WordPress re-emits as the `requests-request.progress` action. It can observe chunks today, but it cannot back a real streaming feature: the full body is still buffered and returned regardless, the transfer cannot be stopped from it, it is a global action with no per-request context so a caller cannot tell which request a chunk belongs to, and it is only clean on cURL. It also cannot feed the SDK, which pulls from a readable body rather than receiving pushed chunks.
- Writing to a custom stream wrapper or `php://output` through `filename`. This does deliver chunks without buffering, but it is a write-only sink with no readable stream to pull from, and it runs entirely inside the blocking call. `WP_Http` also rejects targets like these, since it requires the stream destination to be a writable directory. So it cannot give the SDK a readable body either.
- A pull-based readable body is a larger change than a push callback, since it needs the transport to read in steps rather than in one blocking call. The callback is smaller, but it does not satisfy the SDK's pull model on its own, which is why the readable body is the primary ask.

## Additional context (optional)

- AI SDK streaming PR: https://github.com/WordPress/php-ai-client/pull/255
- The SDK reads the body as a PSR-7 stream and pulls until the end of the stream, so it needs a body that can be read incrementally, not one pushed through a hook.

- [x] I intend to create a pull request to implement this feature myself.

thelovekesh:
Fixes: https://github.com/WordPress/Requests/issues/1056

## Pull Request Type

This is a:
- [ ] Bug fix
- [x] New feature

## Context
Adds opt-in support for reading HTTP response bodies incrementally as they arrive instead of buffering the entire response before returning. This provides the low-level primitive needed for streaming consumers while preserving the existing behavior for all non-streaming requests.

## Detailed Description

- Adds an opt-in streaming mode for Requests responses.
- Introduces a readable response body that can be consumed incrementally while the transfer is still in progress.
- Updates the cURL and fsockopen transports to expose response data without eagerly buffering the complete body when streaming is enabled.
- Preserves the existing buffered behavior by default, so there are no changes for existing callers.
- Lays the foundation for `WP_Http` to expose streaming responses and for higher-level consumers, such as the AI client SDK, to process streamed HTTP responses using a pull-based API.

## Quality assurance

## Documentation

For new features:
- [x] I have added documentation about this feature to the [`docs`](https://github.com/WordPress/Requests/tree/develop/docs) directory.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -25,6 +25,10 @@ Available Hooks
 
     Parameters: `string $data`, `int $response_bytes`, `int|bool $response_byte_limit`
 
+    When the [`stream`](usage-advanced.md#streaming-responses) option is enabled,
+    this hook fires once per `WpOrg\Requests\Response::read()` call as the caller
+    pulls the body, not while the transport receives the response.
+
 * **`requests.before_parse`**
 
     Alter the raw HTTP response before parsing.
@@ -94,6 +98,11 @@ Available Hooks
     When a non-blocking request was made (`$options['blocking'] = false`), both the `$headers` string as well as the `$info` array will be empty.
 
     Prior to Requests 2.1.0, the `$info` parameter was omitted for non-blocking requests.
+
+    When the [`stream`](usage-advanced.md#streaming-responses) option is enabled,
+    this hook fires once the response headers are in, so the `$info` array
+    reflects the transfer at that point (timing and size fields are not final
+    yet).
 
 * **`curl.before_multi_add`**
 
--- a/docs/usage-advanced.md
+++ b/docs/usage-advanced.md
@@ -30,6 +30,67 @@ $session->useragent = 'My-Awesome-App';
 $session->options['useragent'] = 'My-Awesome-App';
 ```
 
+## Streaming Responses
+
+By default, Requests buffers the whole response body in memory before returning.
+For large downloads or long-lived responses, such as Server-Sent Events from an
+AI API, set the `stream` option to `true` instead. The request then returns as
+soon as the response headers are in, and you read the body yourself:
+
+```php
+$response = \WpOrg\Requests\Requests::get('https://example.org/events', array(), array('stream' => true));
+
+// The status code and headers are available right away.
+var_dump($response->status_code);
+
+// Pull the body off the wire in chunks.
+while (!$response->eof()) {
+  $chunk = $response->read();
+  // do something with $chunk
+}
+
+// Release the connection when done.
+$response->close();
+```
+
+The streaming API on `WpOrg\Requests\Response` is small:
+
+- `read($length = 8192)` reads up to `$length` bytes of the body, blocking until
+  data arrives. An empty string means the body has ended.
+- `eof()` tells you whether the body has been fully read.
+- `close()` releases the underlying connection. Also safe to call on responses
+  that are not streamed.
+- `is_streaming()` tells you whether the response is being streamed. Custom
+  transports without streaming support fall back to a normal buffered response,
+  which this lets you detect.
+
+For line-based protocols such as NDJSON or Server-Sent Events, buffer the chunks
+until a newline shows up. See
+[`examples/stream.php`](https://github.com/WordPress/Requests/blob/develop/examples/stream.php)
+for a runnable example.
+
+A few things change when streaming:
+
+- `$response->body` stays empty and `$response->raw` only contains the headers.
+  The body comes out of `read()` instead.
+- Bytes from `read()` are already de-chunked. Both transports ask the server for
+  an uncompressed (`identity`) body, and Requests never decompresses streamed
+  bytes, so what the server sends is what you get.
+- The `timeout` option turns into an idle timeout: the maximum wait for the next
+  chunk of data. A long-lived stream stays open as long as data keeps arriving.
+- `max_bytes` and redirects work as usual. Redirect responses along the way are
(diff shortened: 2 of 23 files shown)


.github/workflows/quicktest.yml
```diff
@@ -20,29 +20,28 @@ jobs:
...
         with:
           php-version: ${{ matrix.php }}
           ini-values: zend.assertions=1, error_reporting=-1, display_errors=On, display_startup_errors=On, log_errors_max_len=0
           coverage: none
 
-      # At least one test needs a non-en_US locale to be available, so make sure it is.
-      - name: Install locales
-        run: |
-          sudo apt-get update
-          sudo apt-get install locales-all
```
thelovekesh, line 45: `locale-gen` is already installed on the GitHub runners (see: https://github.com/actions/runner-images/blob/ubuntu24/20260705.232/images/ubuntu/Ubuntu2404-Readme.md), so there's no need to install it. Running `locale-gen de_DE.UTF-8` is sufficient to generate the required locale.

jrfnl: Please undo this change (here and in the other workflow). Different ubuntu images have different features. We added this in the past because we ran into problems with this before. The fact that it incidentally works without this on Ubuntu 24.04 is irrelevant for long term maintenance of tis workflow and this repo.

.github/workflows/quicktest.yml
```diff
@@ -20,6 +20,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
```
jrfnl: Please remove (here and in the other workflow).

jrfnl:
@thelovekesh Out of curiosity: could this feature be created as an add-on package to Requests instead of having to go into te main package ?

thelovekesh:
@jrfnl We need this in core because it vendors this package directly, so an add-on can't access it. Also, the transport classes are `final`, so an add-on can't extend them and would need to fork them. The part that can live outside already does: the PSR-7/PSR-18 adaptation stays in the existing adapter package.

.github/workflows/quicktest.yml
thelovekesh: While I have reverted it, I only quoted the deleted part in the selected diff (my bad), it was replaced with 
```yml
- name: Generate the de_DE.UTF-8 locale
  run: sudo locale-gen de_DE.UTF-8
```

I removed it because, it was adding extra time per run. With above command, it does the same thing.

Here is the commit. I can re-add if looks good to you - https://github.com/WordPress/Requests/commit/f353f41e6291111d612f3faa0c4818c97ded153c

.github/workflows/quicktest.yml
thelovekesh: Added for testing, reverted now.

jrfnl:
@thelovekesh Hmm... the current team barely has the bandwidth to keep up with regular maintenance and security issues, let alone deal with reviewing and maintaining new features.

.github/workflows/quicktest.yml
jrfnl: No thanks.

PR #1086, opened 2026-09-17
Harden RFC 10025 cookie parsing

### LINKED ISSUE #1084: Review which characters should be `trim`med for cookie and header data
## Setting the stage

By default, the PHP native `[lr]trim()` functions, trim ASCII whitespace and the NUL byte character.
PHP 8.6 changes the default value of the `$characters` parameter to also include the form feed - `"\f"` - character, which was previously not trimmed.

⚠️ _Keep in mind that `[lr]trim()` only operates on the leading and/or trailing characters for a text string. It does not affect the characters in the "middle" !_

Refs:
* https://wiki.php.net/rfc/trim_form_feed
* https://www.php.net/manual/en/function.trim.php

## What has already been done

Requests uses `[lr]trim()` in various places throughout the codebase.
To make this code PHP cross-version compatible, PR #1083 made the following changes:
* Introduced two class constants to represent the different default values for the `$characters` parameter.
* Made the `$characters` being trimmed explicit in each of the `[lr]trim()` function calls (if it wasn't already).

To determine which characters should be trimmed, the following rule of thumb has been used for the initial changes made in PR #1083:
_"Use the PHP 8.6 default (`Trim::WHITESPACE_CHARS`), except when the trimming may be subject to an RFC or other documented rules, in which case use the PHP < 8.6 default (`Trim::WHITESPACE_CHARS_NO_FF`)"_

That way, we preserved existing behaviour in "important" places, while benefitting from the new default value everywhere else.

## Task

For all calls to `[lr]trim()` which currently use the `Trim::WHITESPACE_CHARS_NO_FF` value for `$characters`, it should be researched whether that is correct by checking RFCs and other applicable rules. In some cases, it may also be prudent to evaluate how other projects similar to Requests, like for instance Guzzle, handle these text strings.

If the trimming is correct, it should be documented that this was verified via a comment in this issue.

If a change is necessary, please submit a PR and include tests safeguarding the change to the value passed for `$characters`.

After PR #1083, there will be 14 places in the codebase where this review task is necessary.

asllanmaciel:
## Pull Request Type

This is a:
- [x] Bug fix
- [ ] New feature

## Context

Follow-up to #1084 after #1083. The initial patch narrowed cookie whitespace trimming to protocol WSP (`SP / HTAB`). During review, RFC 10025 was identified as the current cookie specification (it obsoletes RFC 6265) and adds an important first parsing step: a `Set-Cookie` string containing `%x00-08 / %x0A-1F / %x7F` must be ignored.

This update separates those two responsibilities instead of treating all control characters as trim candidates.

Fixes #1084.

## Detailed Description

- renames the internal cookie whitespace constant to `Trim::WHITESPACE_CHARS_RFC10025` and keeps normalization limited to `SP / HTAB`;
- detects the RFC 10025-disallowed control-character ranges before parsing;
- `Cookie::parse()` rejects such direct inputs with `InvalidArgument` instead of silently normalizing them;
- `Cookie::parse_from_headers()` ignores an invalid `Set-Cookie` entry and continues processing the other response cookies, matching the RFC's user-agent parsing behavior;
- adds data-driven coverage for all 32 disallowed control-code points plus mixed valid/invalid response headers;
- keeps HTAB valid and normalized as WSP.

RFC reference: https://www.rfc-editor.org/rfc/rfc10025#section-5.6

## Quality assurance

- [ ] This change does NOT contain a breaking change (invalid cookie strings which were previously accepted are now rejected/ignored as required by RFC 10025).
- [x] The new behavior is covered by focused tests.

Validation on PHP 8.3.14 / PHPUnit 10.5.64:

- RED before implementation: 33/33 new cases failed (32 disallowed CTLs + mixed-header ignore case);
- focused cookie suite after implementation: **165 tests / 440 assertions — PASS**;
- `composer lint`: **176 files — PASS**;
- PHPCS on the three changed files — **PASS**;
- `git diff --check` — **PASS**;
- full suite on this branch: 3,238 tests / 5,486 assertions, 50 failures, 6 warnings;
- clean `develop` baseline (`6200ba9a`): 3,202 tests / 5,410 assertions, the same 50 failures and 6 warnings. The baseline failures are unrelated environment/integration failures (public test-server redirects/proxy availability), so this patch adds no full-suite regressions.

## Documentation

No public API feature is added; the behavior is protocol hardening and is covered inline by the RFC-linked comments/tests.

## AI assistance

AI-assisted tooling was used to investigate the RFC change, prepare the tests/patch, and compare the branch against a clean upstream baseline. The final code, test results, and scope were reviewed directly by the contributor.


masakielastic:
One thing we may want to reconsider before merging this: RFC 6265 has been obsoleted by RFC 10025, and the newer Set-Cookie parsing algorithm is slightly stricter than what this PR currently tests.

RFC 10025 first requires a user agent to reject the entire `set-cookie-string` if it contains:

```text
%x00-08 / %x0A-1F / %x7F
```

i.e. CTLs excluding HTAB. Only after that validation does it remove leading/trailing `WSP` (`SP / HTAB`) from the name/value and attributes.

So the `"Non-WSP control characters are not stripped from cookie values"` test, which currently preserves VT in the parsed value, may need another look: under RFC 10025, VT (`%x0B`) should cause the Set-Cookie string to be rejected rather than preserved.

This also suggests that the hardening may need two separate pieces: WSP trimming (`SP / HTAB`) and CTL validation.

asllanmaciel:
Good catch on RFC 10025. I updated the PR in 9bdaa17b81d8daa7c7850d8386f2af53ac6db2c1: the cookie WSP constant now references RFC 10025, all %x00-08 / %x0A-1F / %x7F controls are rejected by direct Cookie::parse(), and parse_from_headers() ignores only the invalid Set-Cookie entry so other response cookies continue to be processed. I also replaced the VT-preservation expectation with data-driven coverage for all 32 disallowed control values plus a mixed valid/invalid response-header case. The new tests were RED before the change and are GREEN now (165 focused tests / 440 assertions); lint and changed-file PHPCS pass, and the full-suite failure count is identical to a clean develop baseline.

